PR #7, opened 2015-09-19
Carthage enabled

RuiAAPeres:
I added the core + ui (left Alamofire and FLAnimatedImage for now)


kean:
I've added Carthage support 44e5e3b39757ab00b8d34c9b29798b931c1bd995 without copying source files. Also updated readme, CocoaPods setup etc. CocoaPods is still required for demo project, playground, and Alamofire and FLAnimatedImage integration. I'll have to figure out the best way to do it using Carthage.

RuiAAPeres:
@kean :+1:

PR #12, opened 2015-09-27
Update carthage stanza in README

nanoxd:
Carthage cannot parse `'` for strings. It also requires the username/organization that the repo can be found in.


kean:
Right, I just copied it from podfile ;)

Issue #13, opened 2015-09-28
Crash on 32-bit iOS simulator on Mac with 16Gb+ of RAM


RuiAAPeres:
With the following command:

`./xctool -workspace <workspace> -scheme <scheme> -sdk iphonesimulator test`

I am getting this:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   Nuke.Nuke                     	0x004126ac _TFC4Nuke16ImageMemoryCacheCfMS0_FT_S0_ + 588
1   Nuke.Nuke                     	0x003e557a _TIFV4Nuke25ImageManagerConfigurationcFMS0_FT10dataLoaderPS_16ImageDataLoading_7decoderPS_13ImageDecoding_5cacheGSqPS_18ImageMemoryCaching___S0_A1_ + 42
2   Nuke.Nuke                     	0x003f1bd4 globalinit_33_26AEE105B8E103C743A4428E83DD33A7_func0 + 100
3   libdispatch.dylib             	0x032656fd _dispatch_client_callout + 14
4   libdispatch.dylib             	0x032514d5 dispatch_once_f + 83
5   libswiftCore.dylib            	0x02e26fda swift_once + 42
6   Nuke.Nuke                     	0x003f1c64 _TZFC4Nuke12ImageManager6sharedfMS0_FT_PS_13ImageManaging_ + 68
7   Nuke.Nuke                     	0x0040c27d _TF4Nuke21startPreheatingImagesFGSaVS_12ImageRequest_T_ + 29
```

Inside Xcode I was able to reproduce the crash very few times, while on Xctool always crashes. On Xcode the breakpoint hit here: https://github.com/kean/Nuke/blob/44e5e3b39757ab00b8d34c9b29798b931c1bd995/Nuke/Source/Core/ImageMemoryCache.swift#L65

RuiAAPeres:
With the following works:

```
./xctool -workspace ~/Documents/MailOnline/Money/Money.xcworkspace -scheme Money -sdk iphonesimulator -destination name="iPhone 6" test  
```

It seems to me the issue is related to 32bits devices (iPhone5 also crashes, but no the 5s)

RuiAAPeres:
A bit more digging revealed the following:

```
fatal error: floating point value can not be converted to Int because it is greater than Int.max
```

kean:
Fixed by #14

PR #14, opened 2015-09-29
Fixed Crash on 32bits Architecture

RuiAAPeres:
This fixes issue #13

### DIFF
--- a/Nuke/Source/Core/ImageMemoryCache.swift
+++ b/Nuke/Source/Core/ImageMemoryCache.swift
@@ -60,9 +60,12 @@ public class ImageMemoryCache: ImageMemoryCaching {
     
     public class func recommendedCacheTotalLimit() -> Int {
         #if os(iOS)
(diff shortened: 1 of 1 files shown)


kean:
Thanks, I'll take a look a little bit later. I guess that NSProcessInfo.processInfo().physicalMemory returns invalid value like UInt64.max starting with iOS 9 (can't test atm). If it does then we should either find an alternative API or just return some default value as a recommendedCacheTotalLimit. Returning Int.max is unreasonable, 32-bits devices can't even address more memory.

RuiAAPeres:
I tested that on the simulator (my carthage file is currently pointing to that commit) and it's working. But we can look for another alternative.

kean:
Oh, actually physicalMemory works fine. It's just that you have more than 8 Gb of RAM :)

(1024 * 1024 * 1024 * 16) * 0.2 = 3 435 973 836
3 435 973 836 > 2 147 483 647

Issue #35, opened 2015-10-29
Got error message in console: "<Error>: CGBitmapContextCreate: unsupported parameter combination: set CGBITMAP_CONTEXT_LOG_ERRORS environmental variable to see the details"


vkaramov:
I've found the same issue that SDWebImage had. When trying to show image like that http://7xngft.com2.z0.glb.qiniucdn.com/thn.jpg  I've got an error message:

<Error>: CGBitmapContextCreate: unsupported parameter combination: set CGBITMAP_CONTEXT_LOG_ERRORS environmental variable to see the details

After setting CGBITMAP_CONTEXT_LOG_ERRORS environment variable, I got more detailed info:

<Error>: 
	CGBitmapContextCreate: unsupported parameter combination:
	 	16 integer bits/component;
	 	64 bits/pixel;
		RGB color space model; kCGImageAlphaLast;
		1760 bytes/row.
	Valid parameters for RGB color space model are:
		16  bits per pixel,		 5  bits per component,		 kCGImageAlphaNoneSkipFirst
		32  bits per pixel,		 8  bits per component,		 kCGImageAlphaNoneSkipFirst
		32  bits per pixel,		 8  bits per component,		 kCGImageAlphaNoneSkipLast
		32  bits per pixel,		 8  bits per component,		 kCGImageAlphaPremultipliedFirst
		32  bits per pixel,		 8  bits per component,		 kCGImageAlphaPremultipliedLast
		64  bits per pixel,		 16 bits per component,		 kCGImageAlphaPremultipliedLast
		64  bits per pixel,		 16 bits per component,		 kCGImageAlphaNoneSkipLast
		128 bits per pixel,		 32 bits per component,		 kCGImageAlphaNoneSkipLast |kCGBitmapFloatComponents
		128 bits per pixel,		 32 bits per component,		 kCGImageAlphaPremultipliedLast |kCGBitmapFloatComponents
	See Quartz 2D Programming Guide (available online) for more information.

Environment: iOS 9.0 simulator, Xcode 7.1.

kean:
<img width="740" alt="screen shot 2015-10-30 at 21 43 24" src="https://cloud.githubusercontent.com/assets/1567433/10854828/45b00d92-7f4f-11e5-8647-0b3cf30b140f.png">

kean:
64 bpp, 16 bpc, kCGImageAlphaLast (what image data tells us) is not supported
[...]

PR #36, opened 2015-10-29
Fix issue #35

vkaramov:
I've fixed issue #35 by skipping further processing of images with Alpha channel.


kean:
I actually thought about drawing all images in a single format that is best for rendering. I haven't figured out which one to use yet but kCGImageAlphaPremultipliedFirst looks good.

vkaramov:
I have also tried it, but I got another warning ("Unsupported pixel description - 3 components, 64 bits-per-component)

kean:
Ok, thanks a lot. I'll take a closer look when I get home :+1:

kean:
Fixed in 9e24eca381aa5554bc8c78f60068e33ce1e7facc. Releasing v1.0.1 today.

PR #47, opened 2015-12-03
Fix size rounding errors in image downscaling

pyrtsa:
As asked in https://github.com/kean/Nuke/pull/42#issuecomment-160642586.


kean:
Awesome, thanks!

PR #69, opened 2016-05-03
Add BITCODE_GENERATION_MODE to bitcode in Release building to have ca…

vincentsaluzzo:
Add BITCODE_GENERATION_MODE to bitcode in Release building to have capacity to generate bitcode from Carthage build

Refer to this tips in Carthage: <https://github.com/Carthage/Carthage/issues/535>


kean:
Thanks!

PR #131, opened 2017-06-12
Now removing existing node from linked list when adding node with same key

klop:
`Cache.add(_:)` does not check to see if a node already exists for a given key. While the existing node in `map` will be overwritten, it will not be removed from `list`.

This PR checks for an `existingNode` in `map` (where `existingNode.value.key == node.value.key`)  and properly removes it.

### DIFF
--- a/Sources/Cache.swift
+++ b/Sources/Cache.swift
@@ -100,6 +100,10 @@ public final class Cache: Caching {
     }
 
     private func add(node: Node<CachedImage>) {
+        if let existingNode = map[node.value.key] {
+            remove(node: existingNode)
+        }
+        
         list.append(node)
         map[node.value.key] = node
         totalCost += node.value.cost


kean:
It seems that there is also an issue where `totalCost` wouldn't get updated properly if you an item which already exists in cache. It just adds a cost of an added item now

klop:
@kean I was also thinking about the thread-safety of `costLimit` and `countLimit`. Correct me if I'm wrong, but they're the only things preventing `Cache` from having a fully thread-safe public interface. A potential fix would be to have private backing vars for internal use (`_costLimit`, `_countLimit`), while the `costLimit` and `countLimit` implement accessors that lock appropriately. What do you think?

kean:
@klop I think that existing locking inside a `didSet` is sufficient based on the fact that those two properties are simple value types (no reference counting involved). I can't say for sure because documentation on thread-safety either lacking or I'm just not able to find it anywhere.

This is the existing implementation:

```swift
/// The maximum total cost that the cache can hold.
public var costLimit: Int { didSet { lock.sync { trim() } } }

/// The maximum number of items that the cache can hold.
public var countLimit: Int { didSet { lock.sync { trim() } } }
```

The only problem that I see is that if for example you set `costLimit` with two different values simultaneously from two different threads then both `trim` calls might use the same value (the one which was set last). However, the last call is always guaranteed to use the latest value, which is good.

I do have a `testThreadSafety` test (see `CacheTests.swift` file) which tests both of those properties. I can't say for sure how just how efficient it is, but it's good at catching simple threading issues. For example, if I change `costLimit` implementation to this (no lock): `var costLimit: Int { didSet { trim() } }` it would catch an exception pretty quickly (but it might require increasing the iterations count):

Maybe it's better to change it to what you're suggesting just to be safe. It would be a little bit more verbose, but deterministic. What do you think?

klop:
@kean given the tests haven't caught any problems, I think what you've said makes a lot of sense! I think it should be kept as it is for the time being.

PR #137, opened 2017-08-09
Add Swift 4 support

irace:
I can understand not wanting to merge this into `master` until Xcode 9 is out of beta – you may want to make a `swift-4` branch that can become the base branch for this PR?

### DIFF
--- a/Sources/Cache.swift
+++ b/Sources/Cache.swift
@@ -66,7 +66,7 @@ public final class Cache: Caching {
         #endif
     }
 
-    private static func defaultCostLimit() -> Int {
+    public static func defaultCostLimit() -> Int {
         let physicalMemory = ProcessInfo.processInfo.physicalMemory
         let ratio = physicalMemory <= (1024 * 1024 * 512 /* 512 Mb */) ? 0.1 : 0.2
         let limit = physicalMemory / UInt64(1 / ratio)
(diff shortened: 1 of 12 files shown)


Sources/Cache.swift
```diff
@@ -66,7 +66,7 @@ public final class Cache: Caching {
         #endif
     }
 
-    private static func defaultCostLimit() -> Int {
+    public static func defaultCostLimit() -> Int {
```
kean: This things used to be private.

Sources/Cache.swift
```diff
@@ -117,7 +117,7 @@ public final class Cache: Caching {
     }
 
     /// Removes all cached images.
-    public dynamic func removeAll() {
+    @objc public dynamic func removeAll() {
```
kean: I haven't yet went through all of Swift 4 changes. Not sure, why is this change necessary?

Sources/Cache.swift
irace: As of Swift 4, they can’t be private because they’re used as default arguments in public methods (I believe constructors in this specific case).

Sources/Cache.swift
irace: https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md

kean:
Thanks!

kean:
FYI going to release a new version with minimum changes soon which supports both Xcode 8 & 9 and both Swift 3.2 & 4.

PR #146, opened 2018-01-16
Fix disk cache path for macOS

willdahlberg:
From Apple docs:
“In macOS, path is the location at which to store the on-disk cache.

In iOS, path is the name of a subdirectory of the application’s default cache directory in which to store the on-disk cache (the subdirectory is created if it does not exist).”

So this updates the path for macOS accordingly.


kean:
Thanks, @willdahlberg! Totally missed that. Let me know if there are any other issues on macOS or issues in general.

willdahlberg:
Will do. Thank you!

PR #161, opened 2018-05-15
Fix the contentModes not set when initializing a ImageLoadingOptions object

jeromescheer:
Hi, I recently got tired of SDWebImage and wanted something more up to date and Swift based so I added your awesome library (I really like how it's crafted!) in a project but didn't manage to make the `contentModes` attribute work properly. My image placeholder was always displayed as `.scaleAspectFit` instead of `.center`. I think I figured out why.. 😄


jeromescheer:
Great!! Thank you, that was fast! Will do if I see something 😉

PR #163, opened 2018-05-27
Allow custom DataCaching implementation

vmartinelli:
Hello.

First of all, congratulations for the framework and this great new version!

About the PR:

I know that it is planned the disk cache API opening, but I'm using a custom (and private) disk caching mechanism in my app, so I made some changes do allow this.

### DIFF
--- a/Sources/ImagePipeline.swift
+++ b/Sources/ImagePipeline.swift
@@ -576,13 +576,8 @@ public /* final */ class ImagePipeline {
         }
 
         // Store data in data cache (in case it's enabled))
-        if let dataCache = configuration.dataCache, let key = session.request.urlString {
-            if data.isEmpty {
-                dataCache.removeData(for: key)
-            }
-            else {
-                dataCache.storeData(data, for: key)
-            }
+        if !data.isEmpty, let dataCache = configuration.dataCache, let key = session.request.urlString {
+            dataCache.storeData(data, for: key)
         }
 
         _session(session, processFinalImage: image, for: Array(session.tasks.keys))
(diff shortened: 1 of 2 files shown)


kean:
Hey, @vmartinelli 

You're exactly right, one of the goals for Nuke 7 is to bring custom disk caching support. I was planning it for Nuke 7.2. I'll see what I can do to make roadmap public - I'm currently managing it in a private Trello board.

This seems like a great start. I'm almost 100% certain at this point that the API should be asynchronous and it should return Cancellable (but maybe optional? or we could add Cancellable.noOp?).

I'm not yet sure whether to pass any context when storing the responses, haven't come up with any use cases yet. This API (`DataCaching`) is designed for implementing aggressive caching (no HTTP validation needed) which means that we probably don't need to pass URLResponses. There are already ways to implement those types of sophisticated custom caches either via `Nuke.DataLoading` protocol, or `Foundation.URLProtocol`.

I think I can merge this in a way it is and then continue refining it before the released - add tests, documentation, etc.

It seems there is a flaky preheater test, I'm currently working on them in master.

Sources/ImagePipeline.swift
```diff
@@ -575,8 +576,13 @@ public /* final */ class ImagePipeline {
         }
 
         // Store data in data cache (in case it's enabled))
-        if !data.isEmpty, let dataCache = configuration.dataCache, let key = session.request.urlString {
-            dataCache[key] = data
+        if let dataCache = configuration.dataCache, let key = session.request.urlString {
+            if data.isEmpty {
+                dataCache.removeData(for: key)
```
kean: I don't think that you should remove data at this point. If you get to this point in the pipeline it means that the data wasn't found in the `dataCache` and was fetched from the network instead.

vmartinelli: You're right. In addition, the data removal logic must be in the cache mechanism itself. The changes have been made.

kean:
Thanks for updating the PR. I've merged it, but please keep in mind that I'll might change the API a little bit.

PR #169, opened 2018-06-16
Update Install Version with CocoaPod from 6.0 to 7.0

Alxegore:


kean:
Thanks, I really appreciate fixes in docs / guides!

Issue #174, opened 2018-07-05
UITableViewCell image mix-up


michaelnisi:
Looks like I’ve found an interesting edge case, provoking Nuke to load the wrong image into a table view cell.

![img_1237](https://user-images.githubusercontent.com/108903/42304168-2acbba74-8025-11e8-893c-8b42b37216ed.jpg)

The loading code, called from within the table view cell, does nothing special really:

```swift
  private static func load(
    url: URL,
    into view: UIImageView?,
    sized size: CGSize,
    cb: @escaping ImageTask.Completion
  ) {
    let req = ImageRequest(
      url: url,
      targetSize: size,
      contentMode: .aspectFill
    ).processed(with: ScaledWithRoundedCorners(size: size))

    guard let v = view else { return }

    os_log("loading image: %{public}@ %{public}@", log: log, type: .debug,
           url as CVarArg, size as CVarArg)

    guard let currentImage = v.image else {
      Nuke.loadImage(with: req, into: v, completion: cb)
      return
    }

    os_log("placeholding with current image", log: log, type: .debug)

    let opts = ImageLoadingOptions(
      placeholder: currentImage,
      transition: nil,
      failureImage: currentImage,
      failureImageTransition: nil,
      contentModes: nil
    )

    Nuke.loadImage(with: req, options: opts, into: v, completion: cb)
  }
```

It’s solid for all use cases, except for one. Erratically, **after state restoration in landscape mode**, my table view appears with mixed up images. Scrolling, of course, fixes the problem immediately.

My pipeline mainly uses defaults:

```swift
  init() {
    let pipeline = ImagePipeline {
      $0.imageCache = ImageCache.shared
      
      let config = URLSessionConfiguration.default
      config.urlCache = nil
      $0.dataLoader = DataLoader(configuration: config)
      
      $0.dataCache = try! DataCache(name: "ink.codes.podest.images") { name in
        return String(djb2Hash32(string: name))
      }
      
      $0.dataLoadingQueue.maxConcurrentOperationCount = 6
      $0.imageDecodingQueue.maxConcurrentOperationCount = 1
      $0.imageProcessingQueue.maxConcurrentOperationCount = 2

      $0.isDeduplicationEnabled = true

      $0.isProgressiveDecodingEnabled = false
    }

    ImagePipeline.shared = pipeline
  }
```

[...]

PR #190, opened 2018-09-30
Reset task ID

michaelnisi:
### Tracking Down #174

Of course, state restoration in landscape mode is mere correlation, but the related selecting of the current table view row exposed itself causing #174.

```swift
selectRow(at:animated:scrollPosition:) - UITableView
```

Investigating further showed, the issue only occurred with quick synchronous memory cache lookup, when `imageCache.cachedResponse(for: request)` returned a response. 

A comment, in `ImageView.swift`, made me curious.

```swift      
// Make sure that view reuse is handled correctly.
self.taskId += 1
let taskId = self.taskId
```

And sure enough, resetting the task identifier after an image cache hit, resolves this issue. 🙌

Unfortunately, I couldn’t come up with a tight test case, reproducing this timing issue. Maybe you can think of way to test this.

Sidebar, why not use `UUID` as `taskId`, instead of incrementing?

### DIFF
--- a/Sources/ImageView.swift
+++ b/Sources/ImageView.swift
@@ -321,6 +321,9 @@ private final class ImageViewController {
         if request.memoryCacheOptions.isReadAllowed,
             let imageCache = pipeline.configuration.imageCache,
             let response = imageCache.cachedResponse(for: request) {
+
+            self.taskId = 0
+
             handle(response: response, error: nil, fromMemCache: true, options: options)
             completion?(response, nil)
             return nil


michaelnisi:
I should have established first, `Nuke.loadImage(with:, options:, into:, completion:)` happens during `UITableViewCell` configuration in the data source, nothing special.

kean:
Hey, spot on! Thanks a lot for the investigation and for the fix.

The issue is extremely hard to reproduce. In this scenario the outstanding request still gets canceled but because cancellation is asynchronous in `ImagePipeline` there is a small window where it still might call a completion handler for a canceled task. I think it might be possible to add a guarantee that the completion won't be called.

I've written a test spec that describes this behavior. Sure enough, it's red and with your fix it becomes green.

```swift
func testDisplayedCachedImageWhenThereIsPendingRequest() {
    mockPipeline.queue.isSuspended = true

    let requestA = ImageRequest(url: URL(string: "test://imageA")!)
    let requestB = ImageRequest(url: URL(string: "test://imageB")!)

    // Given pipeline with cancellation disabled (important!)
    mockPipeline.isCancellationEnabled = false

    // Given an image A not stored in cache and image B - stored.
    let imageB = Image()
    mockCache[requestB] = imageB

    // Given an image view which is in the process of loading the image A.
    Nuke.loadImage(with: requestA, into: imageView) { _, _ in
        // Expect completion to never get called, we're already displaying
        // the image B by that point.
        XCTFail("Enexpected completion for requestA")
    }

    // When starting a starting a new request for the image B.
    Nuke.loadImage(with: requestB, into: imageView)

    // Expect an image B to be displayed immediatelly.
    XCTAssertEqual(imageB, imageView.image)

    // When the pipeline finishes loading the image B.
    expectNotification(MockImagePipeline.DidFinishTask)
    mockPipeline.queue.isSuspended = false
    wait()

    // Expect an image view to still be displaying the image B.
    XCTAssertEqual(imageB, imageView.image)
}
```

I'm going to add this test - and maybe a few more - once it's merged.

kean:
@michaelnisi I was adding more tests and found a similar scenario with the same issue. Calling `Nuke.cancelRequest(for:)` method also had the same flaw, I fixed it.

michaelnisi:
So, in that case, `Nuke.cancelRequest(for:)` would cancel the wrong request? 
💯 I’m glad, we were able to resolve these. Thank you for your responsiveness.

kean:
> So, in that case, Nuke.cancelRequest(for:) would cancel the wrong request?

It would cancel the correct request, but it would reset request identifier managed by the ImageView. In these rare cases when canceling the task which is "almost" completed, the view it would still display an image.

PR #191, opened 2018-10-03
Warning for transition issue

Nicholas714:
This could have saved me a lot of time debugging. 

The animations would cause the app to exceed the keyboard extension's RAM usage limit when scrolling too fast. With each image animation playing for 0.33 seconds, I believe UIKit was keeping a reference to the UIImage until the animation is complete, preventing Nuke from disposing of the images to free up RAM. So, each image would be kept for at least 0.33 seconds. If you scroll fast during this time, the RAM usage would increase dramatically and be unable to be reclaimed even after removing all images until the animations were complete.

Once the animation was removed, I was able to go from about 120 MB of usage to only 20 MB of usage when scrolling through a collection view of images. This fixed the app's constant crashing due to exceeding the app extension's RAM limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath:
 
 #### Placeholders, Transitions and More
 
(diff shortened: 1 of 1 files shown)


kean:
Thanks, @Nicholas714! I would also add a more details explanation in a Performance Guide.

PR #197, opened 2018-10-14
allow custom processor through init

drkibitz:
- This allows shared custom processors to be provided rather than depending only on factory methods of image requests

e.g.
```swift
let imageRequest = ImageRequest(url: url, processor: MyProccessor.shared)
```

I also added tests for this and my previous pull request for upscaling.

### DIFF
--- a/Sources/ImageRequest.swift
+++ b/Sources/ImageRequest.swift
@@ -146,17 +146,37 @@ public struct ImageRequest {
 
     #if !os(macOS)
 
-    // Convenience initializers with `targetSize` and `contentMode`. The reason
-    // why those are implemented as separate init methods is to take advantage
-    // of memorized `decompressor` when custom parameters are not needed.
+    // Convenience initializers with custom processor. The reason why these
+    // are implemented as separate init methods is to take advantage of
+    // memorized `decompressor` when a custom processor is not needed.
+
+    /// Initializes a request with the given URL.
+    /// - parameter processor: Custom image processer.
+    public init<Processor: ImageProcessing>(url: URL, processor: Processor) {
+        self.init(url: url)
+        self.processor = AnyImageProcessor(processor)
+    }
+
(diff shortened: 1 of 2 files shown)


kean:
Hey, @drkibitz! Thanks, I think it makes sense to add a convenience initializer like that.

I would suggest making this init generic to hide `AnyImageProcessor` which exists because of the Swift limitations:

```swift
public init<Processor: ImageProcessing>(urlRequest: URLRequest, processor: Processor) {
        self = ImageRequest(urlRequest: urlRequest)
        self.processor = AnyImageProcessor(processor)
 }
```

kean:
```
// Convenience initializers with custom processer. The reason
// why those are implemented as separate init methods is to take advantage
// of memorized `decompressor` when custom parameters are not needed.
```

I'm not sure I follow, could you please clarify what you meant by this comment? I don't see how these new initializes allow you to do something that the existing ones don't.

drkibitz:
@kean That comment was taken from what was already there, changing `"Convenience initializers with targetSize and contentMode."` to `"Convenience initializers with custom processer."`. I didn't mean anything by the comment other than just trying to repeat the pattern of the previous comment but for the custom processor parameter.

drkibitz:
@kean Out of curiosity, what are you gaining by making the image processor objects in this library structs?

drkibitz:
@kean also, FYI, I've updated the PR hopefully to your liking. I've taken a little liberty on the comment(s), and the init calling pattern.

kean:
Oh, right, there is an existing comment like that 😄. I will take a look, not sure why I thought it was necessary, seems outdated. It's there just for convenience and to make API for resizing images - which is a common operation - easier to find and use.

> Out of curiosity, what are you gaining by making the image processor objects in this library structs?

AFAIR to reduce the number of allocations and deallocations, that's it.

PR #201, opened 2018-11-15
alwaysTransition flag

gabzsa:
Added always transition flag to ImageLoadingOptions, so that you can control whether to show transition even when image is retrieved from cache


kean:
@gabzsa thanks, looks good. I checked the CI report, the build is failing on macOS:

```
❌  /Users/travis/build/kean/Nuke/Sources/ImageView.swift:396:66: missing argument for parameter #5 in call
            _display(image, options.transition, fromMemCache, nil)

❌  /Users/travis/build/kean/Nuke/Sources/ImageView.swift:398:85: missing argument for parameter #5 in call
            _display(failureImage, options.failureImageTransition, fromMemCache, nil)
                 ^
❌  /Users/travis/build/kean/Nuke/Sources/ImageView.swift:405:55: missing argument for parameter #5 in call
        _display(image, options.transition, false, nil)
```

PR #205, opened 2018-11-27
Add static library target. (iOS)

eonil:
This PR adds static library target for iOS.

Static library has several benefits when compared to dynamic library (that is in ".framework"). 

- *Faster app load time* due to elision of run-time symbol resolution of dynamic library.
- Library code will be merged into app main executable. Therefore file management becomes simpler.
- Potentially smaller binary due to direct linking. (stripping away unused symbols)
- Potentially better chance for optimization by compiler due to direct linking.

The only disadvantage to ".framework" is it can contain only code symbols, no resource files. But it's not a problem for Nuke library. If I don't need any resources, I always prefer static library in Swift.

### DIFF
--- a/Nuke.xcodeproj/project.pbxproj
+++ b/Nuke.xcodeproj/project.pbxproj
@@ -187,6 +187,17 @@
 		0CF4DE7E1D412A9E00170289 /* ImagePreheater.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CF4DE7C1D412A9E00170289 /* ImagePreheater.swift */; };
 		0CF4DE7F1D412A9E00170289 /* ImagePreheater.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CF4DE7C1D412A9E00170289 /* ImagePreheater.swift */; };
 		0CF4DE801D412A9E00170289 /* ImagePreheater.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0CF4DE7C1D412A9E00170289 /* ImagePreheater.swift */; };
+		453CA5BB21ACEC5C00DD8E81 /* ImageView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C0FD5DF1CA47FE1002A78FB /* ImageView.swift */; };
+		453CA5BC21ACEC5C00DD8E81 /* ImageRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C0FD5D91CA47FE1002A78FB /* ImageRequest.swift */; };
+		453CA5BD21ACEC5C00DD8E81 /* ImagePipeline.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C0FD5D31CA47FE1002A78FB /* ImagePipeline.swift */; };
+		453CA5BE21ACEC5C00DD8E81 /* ImageTaskMetrics.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0C367A9420B9D0D0002342A5 /* ImageTaskMetrics.swift */; };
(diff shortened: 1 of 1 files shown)


eonil:
Test failed for some timeout. How can I perform test again?

kean:
Hey, @eonil. Thanks for the PR! I’m on vacation and don’t have my laptop with me. Will go through the PRs when I’m back.

kean:
Hey, @eonil! Just wanted to let you know that I haven't forgot about this PR.

My concern is that with so many new targets the project becomes a little unwieldy. I'm currently testing a new approach in [FutureX](https://github.com/kean/FutureX) where I have a single multi-platform target (instead of four, one per platform) and where I use [xcconfig](https://github.com/kean/FutureX/blob/master/Supporting%20Files/Future.xcconfig) to extract some of the configuration. As soon as I'm confident that it works, I'm going to try adding a target that builds static libraries. With this new approach it should be nice and easy.

eonil:
@kean Thanks for note.

I have some opinions for your new approach.

1. I don't think "single binary for multiple platforms" would work. Simply because different platforms needs different executable code and packaging. It has many potential problems that wouldn't be visible immediately, but can crash build silently or explicitly. And in that case, it's very hard to fix such problems.This kind of configuration was been used as a workaround 5-6 years ago before Xcode didn't support proper library target for iOS. (no Swift, no module, no framework). One of the biggest problem was Simulator. There's no real good configuration for Simulator. It's your freedom to test such configuration, but  I wouldn't go there unless Apple officially recommends it.

It's okay to close this ticket if you add static library target in any form.


 IMO, framework target is unnecessary for this kind (code-only) of library. Static library is always a better choice.

eonil:
@kean Thanks for note.

I have some opinions for your new approach.

1. I don't think "single binary for multiple platforms" would work. Simply because different platforms needs different executable code, linking and packaging. It has many potential problems that wouldn't be visible immediately, but can crash build silently or explicitly. And in that case, it's very hard to fix such problems. AFAIK, Universal Binary is all about providing "multiple architecture for one platform", rather than multi-platform support.

This kind of configuration was been used as a workaround 5-6 years ago before Xcode didn't support proper library target for iOS. (no Swift, no module, no framework). There was many experiences with exotic configurations, and one of the biggest problem was Simulator. It was impossible to provide proper build settings for Simulator. Even it worked at first, it usually failed when it's been embedded into another large project.

I wouldn't recommend or go there unless Apple officially recommends it.

2. `.xcconfig` can be a good option if you're using many custom build settings. Though I think more custom build settings means less stable build on Apple platforms.

3. IMO, framework target is unnecessary for this kind (code-only) of library. With Swift, there's no reason to make libraries as frameworks unless it needs non-code resources. IMO, it's better to erase all framework targets and keep only static library targets. It's just one target for each platforms.

kean:
> I don't think "single binary for multiple platforms" would work.

It doesn't build a single binary, there are still four different binaries produced, depending on which platform you build for.

kean:
If you’d like to see an example of such target setup, you can see RxSwift and PromiseKit. The latter has been using it for a while, seems that it’s working well for them.

Good point about frameworks, it might be worth considering deprecating them, need to do some research.

gabors:
Don't kill the frameworks pls :)
I love Carthage

kean:
Don't worry, I'm committed to supporting all of the installation methods, including Carthage and manual installation. All Nuke and FutureX satellite projects [use Carthage](https://github.com/FutureXCommunity/CocoaFx).

gabors:
@kean perfect, thanks ;)

kean:
Sorry, this PR is outdated, going to close it for now. I just release Nuke 7.6 which consolidated support for all four platforms in a single target. Now adding a static library becomes viable. I'm not entirely sure I would like to maintain this separate target though.

PR #210, opened 2019-02-20
Add successImageRenderingMode to ImageLoadingOptions

### LINKED ISSUE #209: Need support for UIImage.withRenderingMode(_:)
It would be massively helpful to provide a mechanism for displaying an image with a rendering mode (e.g. `.alwaysTemplate`). PR incoming.

clayellis:
Fixes #209 

@kean Let me know what you think of this implementation. 

I added `successImageRenderingMode` to `ImageLoadingOptions` to provide the mechanism for specifying the rendering mode. It was only needed on the loaded (success) image since both the placeholder and failure images can be provided with a rendering mode already applied.

The rendering mode is applied as high in the chain as possible (in `handle(response:...)` and `handle(partialImage:...)`) in order to avoid propagating more state down into the `_display(...)` methods (since the rendering mode is only applicable to the loaded image).

I added two tests but wasn't able to run them since they're in a `#if !os(macOS)` block — I couldn't figure out how to trigger those tests. I figured I'd write the tests and then ask for your input.


clayellis:
Looks like the test I wrote to check if the rendering mode was successfully applied is failing. It could be a bad test — I'm open to a better way to checking that if you have any suggestions @kean. I can verify outside of a test that it's working as intended, just need a better way to actually test it.

kean:
The code looks good. I ran the tests locally, they succeed. Weird.

clayellis:
Interesting. For my own curiosity, how do you run the test suite targeting iOS so that the iOS-specific tests run?

kean:
Locally I select a "Nuke iOS" scheme and run tests, that's about it. On CI the tests are executed based of the Travis [config file](https://github.com/kean/Nuke/blob/master/.travis.yml).

P.S. in `develop` branch I got rid of per-platform targets. Haven't merged/released it yet. There is now a single `Nuke` target instead. Same with unit tests.

kean:
OK I found there reason why the tests are [failing for tvOS](https://travis-ci.org/kean/Nuke/jobs/496228491). The new `alpha.png` file was not added to  the `Nuke tvOS Tests` target - again, wouldn't have happened in `develop` branch where there is a single target for all platforms.

Not sure what's the issue with [iOS 9.3 yet](https://travis-ci.org/kean/Nuke/jobs/496228489). I don't have it locally, downloading it.

kean:
> Not sure what's the issue with [iOS 9.3 yet](https://travis-ci.org/kean/Nuke/jobs/496228489). I don't have it locally, downloading it.

It seems that you can't use `XCAssertEqual` on `UIImage` objects in iOS 9, we need to find a different approach.

clayellis:
Great, thanks for the insight and time you put into checking on those failures. I'll look into fixes and workarounds.

kean:
Hey, @clayellis? Are you going to follow up on this? If not, I can fix these issues and merge it myself. If you no longer need these feature, please close the MR.

clayellis:
@kean My apologies, I have a task in my backlog to come back to this. Will fix.

clayellis:
@kean I'm not going to be able to get around to fixing that test any time soon.

nickmshelley:
Looks like iOS 9 support was dropped recently, so the test failures should no longer be a problem.

kean:
I'm going to close this PR for now. Please consider updating and reopening it in case it is still relevant.

PR #217, opened 2019-03-27
Update/swift 5

DanielStormApps:
## Summary
- Update for Swift 5 and Xcode 10.2 support

## Notes
- The new `Hashable` updates are outside my comfort zone. There may be a better way to satisfy the new requirements.
    - Seems to me the new `hash(into:)` is a replacement for `hashValue`: https://developer.apple.com/documentation/swift/hashable/2995575-hash

### DIFF
--- a/Sources/ImageRequest.swift
+++ b/Sources/ImageRequest.swift
@@ -253,7 +253,7 @@ public struct ImageRequest {
 public extension ImageRequest {
     /// Appends a processor to the request. You can append arbitrary number of
     /// processors to the request.
-    public mutating func process<P: ImageProcessing>(with processor: P) {
+    mutating func process<P: ImageProcessing>(with processor: P) {
         guard let existing = self.processor else {
             self.processor = AnyImageProcessor(processor)
             return
@@ -264,7 +264,7 @@ public extension ImageRequest {
 
     /// Appends a processor to the request. You can append arbitrary number of
     /// processors to the request.
-    public func processed<P: ImageProcessing>(with processor: P) -> ImageRequest {
+    func processed<P: ImageProcessing>(with processor: P) -> ImageRequest {
         var request = self
         request.process(with: processor)
         return request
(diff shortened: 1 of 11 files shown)


Sources/Internal.swift
```diff
@@ -558,13 +558,14 @@ extension String {
...
     var sha1: String? {
-        guard let input = self.data(using: .utf8) else {
-            return nil
-        }
-        var hash = [UInt8](repeating: 0, count: Int(CC_SHA1_DIGEST_LENGTH))
-        input.withUnsafeBytes {
-            _ = CC_SHA1($0, CC_LONG(input.count), &hash)
+        guard let input = self.data(using: .utf8) else { return nil }
+        
+        let hash = input.withUnsafeBytes { (bytes: UnsafeRawBufferPointer) -> [UInt8] in
```
DanielStormApps: @kean would you prefer an `#if compiler(>=5.0)` here?

kean: Please, that would be ideal. I intend to keep Swift 4.2 support for the next year or so.

Sources/ImageRequest.swift
```diff
@@ -264,34 +264,34 @@ public extension ImageRequest {
...
 internal extension ImageRequest {
     struct CacheKey: Hashable {
         let request: ImageRequest
-
-        var hashValue: Int {
+        
+        func hash(into hasher: inout Hasher) {
             if let customKey = request._ref.cacheKey {
-                return customKey.hashValue
+                hasher.combine(customKey)
```
kean: It seems that this changes the behavior of `CacheKey`, please add return or `else`.

kean:
Thanks you, @DanielStormApps! I reviewed your changes and made a couple of comments.

Also `.travis.yml` needs an update.

Sources/Internal.swift
DanielStormApps: Updated. Added `#if swift(>=5.0)` check.

Sources/ImageRequest.swift
DanielStormApps: Whoops! Good catch. Also updated `LoadKey`, had the same issue.

DanielStormApps:
> Thanks you, @DanielStormApps! I reviewed your changes and made a couple of comments.
> 
> Also `.travis.yml` needs an update.

Updated. Not sure if I applied the changes you had intended. Seems the builds are failing except for 10.2 environments.

kean:
@DanielStormApps merged this plus merged the changes to remove per-platform targets. So now instead of `Nuke iOS`, `Nuke macOS`, etc targets we have a single target. No duplication.

PR #219, opened 2019-04-03
Swift PM: Add tools version to enable Xcode build

shpakovski:
Hi Alexander, it looks like your `master` branch is ready for tools v5. Proposed changes make a package working on Mac. Thanks!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,12 @@
+// swift-tools-version:5.0
 import PackageDescription
 
 let package = Package(
     name: "Nuke",
-    exclude: ["Tests"] // excluding until all module layouts get supported
+    products: [
+        .library(name: "Nuke", targets: ["Nuke"]),
+    ],
+    targets: [
+        .target(name: "Nuke", path: "Sources")
+    ]
 )


kean:
Hi, @shpakovski! This looks good, have you already tested this new package description? Is it possible to keep support for 4.2 packages? I'm not sure anyone is using it though.

shpakovski:
> This looks good, have you already tested this new package description?

Sure, I use exactly this `Package.swift` in my project.

> Is it possible to keep support for 4.2 packages?

Could you please explain why this matters? We’re two months away from Xcode 11, and your codebase is totally future-proof. This library is more ready for production than competition, I suggest to keep it this way and “force-push” customers to use Swift 5.

> I'm not sure anyone is using it though.

Do you mean older Swift or the Swift PM itself, sorry?

kean:
Alright, I think it's safe to assume that people who are ahead of the curve and are already using SwiftPM have also already migrated to Swift 5 as well. If there is someone who can't migrate just yet, they can keep using the previous version of Nuke.

> Do you mean the older version of Swift or the Swift PM itself, sorry?

Sorry, I wasn't clear enough, I meant SwiftPM.

I'm merging this PR, thanks for putting it together!

shpakovski:
I can delete own fork now, thanks!

PR #220, opened 2019-04-08
Swift PM: Include platform for code compatibility

shpakovski:
Hey Alexander, it’s me again. Other changes in 7.6 require macOS 10.12, but Xcode uses 10.10 by default. This is a small fix which helps to get rid of all compiler errors related to modern APIs. Thanks!


kean:
Hi! Thanks, will release a patch some time tomorrow.

Issue #224, opened 2019-04-19
Preheating combined with image loading produces access races


michaelnisi:
Preheating combined with image loading produces access races because locks are accessed from different threads. TSan traps `_createSession` and `loadImage` accessing the `ImagePipeline` lock from different threads. These are the two methods in `ImagePipeline`.

```swift
    private func getNextTaskId() -> Int {
        os_unfair_lock_lock(&lock)
        defer { os_unfair_lock_unlock(&lock) }
        nextTaskId += 1
        return nextTaskId
    }

    private func getNextSessionId() -> Int {
        os_unfair_lock_lock(&lock)
        defer { os_unfair_lock_unlock(&lock) }
        nextSessionId += 1
        return nextSessionId
    }
```

```
WARNING: ThreadSanitizer: Swift access race (pid=9524)
  Modifying access of Swift variable at 0x7b4000006ed8 by thread T9:
    #0 $s4Nuke13ImagePipelineC13getNextTaskId33_64FA2D582EFB9D605832F3421CEE67ADLLSiyF <null> (Nuke:x86_64+0x1f327)
    #1 $s4Nuke13ImagePipelineC04loadB04with8progress10completionAA0B4TaskCAA0B7RequestV_yAA0B8ResponseCSg_s5Int64VAPtcSgyAN_AC5ErrorOSgtcSgtF <null> (Nuke:x86_64+0x1e72e)
    #2 $s4Nuke14ImagePreheaterC16_startPreheating33_337920196C15B3DB24A28FE684F1274ELL4withyAA0B7RequestV_tFyyyccfU_ <null> (Nuke:x86_64+0x9ad10)
    #3 $s4Nuke14ImagePreheaterC16_startPreheating33_337920196C15B3DB24A28FE684F1274ELL4withyAA0B7RequestV_tFyyyccfU_TA <null> (Nuke:x86_64+0x9e880)
    #4 $s4Nuke9OperationC5startyyF <null> (Nuke:x86_64+0x7e327)
    #5 $s4Nuke9OperationC5startyyFTo <null> (Nuke:x86_64+0x7e580)
    #6 __NSOQSchedule_f <null> (Foundation:x86_64+0x97f5a)
    #7 _dispatch_client_callout <null> (libdispatch.dylib:x86_64+0x3db4)

  Previous modifying access of Swift variable at 0x7b4000006ed8 by thread T3:
[...]

PR #225, opened 2019-04-19
Introduce synchronizing queue

michaelnisi:
We are missing a point of synchronization in `ImagePipeline`. This becomes apparent when preheating and image loading are used in combination, as described in #224.

This PR just adds a queue for synchronising access, which resolves this issue. @kean probably knows a better point in the system for controlling access.


kean:
Fixed in https://github.com/kean/Nuke/releases/tag/7.6.2

PR #266, opened 2019-06-24
Draw border around rounded images

michaelnisi:
For some use cases it makes sense to bake borders directly into the image. These changes add an optional border to `ImageProcessing.RoundedCorners`. I hope I’m not the only one who could use something like this.

### DIFF
--- a/Sources/ImageProcessing.swift
+++ b/Sources/ImageProcessing.swift
@@ -174,30 +174,51 @@ extension ImageProcessor {
 // MARK: - ImageProcessor.RoundedCorners
 
 extension ImageProcessor {
-
+  
     /// Rounds the corners of an image to the specified radius.
     ///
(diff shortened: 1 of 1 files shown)


kean:
Hey, @michaelnisi. I got sick, was not opening GitHub for a while.

This seems useful, I've seen similar solutions in other frameworks. There is probably a way to make these processing operations composable. But I will be OK merging it as is and revising this later.

michaelnisi:
Merging this would be cool. Let me go through it one more time, I’ll ping you. Get well 🍵💐

michaelnisi:
@kean I’ve quickly reviewed these changes and after using this branch for some time now, I’d say merging them is fine. Border width gets ignored yet, but I’ve included it for having a stable API. lgtm

PR #270, opened 2019-07-05
Copy edit

michaelnisi:
These are some suggestions for the copy, after I’ve finally took the time to actually read the README. 😅 Pick what you like.

And congrats for the quality, it’s really good. 🥇


kean:
Thank you, @michaelnisi! I really appreciate the MRs like that, I tend to make lots of mistakes in docs! 🍻

PR #291, opened 2019-08-09
Target system queues

michaelnisi:
These changes configure access queues to target system queues. I have written about this [recently](https://troubled.pro/2019/07/mutex.html), it is to control thread creation. 

The quality of service configuration I’m suggesting performs particularly well in my setup. I have also tried setting QoS for Nuke’s operation queues, but didn’t accomplish any palpable performance gains.

### DIFF
--- a/Sources/DataCache.swift
+++ b/Sources/DataCache.swift
@@ -86,7 +86,7 @@ public final class DataCache: DataCaching {
     private let lock = NSLock()
     private var staging = Staging()
 
-    /* testable */ let wqueue = DispatchQueue(label: "com.github.kean.Nuke.DataCache.WriteQueue")
+    /* testable */ let wqueue = DispatchQueue(label: "com.github.kean.Nuke.DataCache.WriteQueue", target: .global(qos: .utility))
 
     /// A function which generates a filename for the given key. A good candidate
     /// for a filename generator is a _cryptographic_ hash function like SHA1.
--- a/Sources/ImagePipeline.swift
+++ b/Sources/ImagePipeline.swift
@@ -21,7 +21,7 @@ public /* final */ class ImagePipeline {
     public let configuration: Configuration
 
     // This is a queue on which we access the sessions.
-    private let queue = DispatchQueue(label: "com.github.kean.Nuke.ImagePipeline")
+    private let queue = DispatchQueue(label: "com.github.kean.Nuke.ImagePipeline", target: .global(qos: .userInteractive))
 
     private var tasks = [ImageTask: TaskSubscription]()
 
--- a/Sources/ImagePipeline.swift
+++ b/Sources/ImagePipeline.swift
@@ -21,7 +21,7 @@ public /* final */ class ImagePipeline {
     public let configuration: Configuration
 
     // This is a queue on which we access the sessions.
-    private let queue = DispatchQueue(label: "com.github.kean.Nuke.ImagePipeline", target: .global(qos: .userInteractive))
+    private let queue = DispatchQueue(label: "com.github.kean.Nuke.ImagePipeline", target: .global(qos: .userInitiated))
 
     private var tasks = [ImageTask: TaskSubscription]()
 
--- a/Sources/ImagePreheater.swift
+++ b/Sources/ImagePreheater.swift
@@ -14,7 +14,7 @@ import Foundation
 /// All `Preheater` methods are thread-safe.
 public final class ImagePreheater {
     private let pipeline: ImagePipeline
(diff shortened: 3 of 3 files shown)


michaelnisi:
Not sure about `.utility` for WriteQueue, though.

kean:
Thanks, @michaelnisi. This looks good, I wanted to experiment with QoS for a while now.

> suggesting performs particularly well in my setup

Did you measure the performance gains?

Sorry, I'm a bit slow to respond because I'm currently focusing on another project.

michaelnisi:
Measuring against master, I’m perplexed how my observations misguided me once again. **master** without setting QoS wins. 

### master
Screenshot 2019-08-12 at 09 37 22

### target
Screenshot 2019-08-12 at 09 37 36

Performance, like weather, is hard to measure—slower but *feels like* faster.

kean:
I'm personally still a bit confused about QoS in GCD. I can see at least one potential use case. `DataCache` has in-memory staging area* and we don't need the data to be written to disk as quickly as possible, so we might want de-prioritize `DataCache` queues.

\* I'm actually a bit skeptical about the staging area itself. I saw clear performance when it was introduced. But I need to test it again, because it might be that it's useless/harmful on AFPS.

michaelnisi:
Trying an explanation, I’m setting QoS throughout my app now as well, this may give Dispatch enough context to perform better overall. Intelligent systems like [GCD](https://developer.apple.com/documentation/DISPATCH) or [V8](https://v8.dev) that adjust dynamically to context and environment are tricky to reason about.

kean:
Yes, it's very tricky to reason about and even trickier to measure. In order for prioritization to have an effect, you need to load the system in a specific way.

michaelnisi:
I have just counted threads, 10 for both.

kean:
> so we might want de-prioritize DataCache queues.

Which you've already done, obviously. By the way, are you measuring with `DataCache` enabled? I don't see data cache entires in a list of signposts.

michaelnisi:
I use HTTP caching, all defaults for the pipeline. Something like this:

```swift
  private static func makeDataLoader() -> DataLoading {
    let conf = DataLoader.defaultConfiguration
    let dice = Int.random(in: 1...6)
    
    if dice == 6 {
      os_log("using protocol cache policy", log: log)
      
      conf.requestCachePolicy = .useProtocolCachePolicy
    } else {
      conf.requestCachePolicy = .returnCacheDataElseLoad
    }
    
    return DataLoader(configuration: conf)
  }

  /// Returns a new image pipeline.
  ///
  /// Sporadically follows `.useProtocolCachePolicy`, instead of 
  /// `.returnCacheDataElseLoad`, for keeping up-to-date. This compromise can be
  /// unfortunate for users offline during this session.
  ///
  /// - Parameter removing: Clears the cache (for development).
  private static func makeImagePipeline(removing: Bool = false) -> ImagePipeline {
    let name = "ink.codes.feedkit.images"
    
    if removing {
      do {
        os_log("trying to remove cache", log: log, type: .info)
        try removeAllFromCache(named: name)
      } catch {
        os_log("nothing to remove", log: log)
      }
    }
    
    ImagePipeline.Configuration.isSignpostLoggingEnabled = true
    
    var conf = ImagePipeline.Configuration()
    conf.dataLoader = makeDataLoader()
    
    return ImagePipeline(configuration: conf)
  }
```

michaelnisi:
👉 Here’s the [use case](https://github.com/michaelnisi/feedkit/blob/master/FeedKit/images/ImageRepository.swift)

michaelnisi:
Note that these measurements are internal, the signposts are placed inside of Nuke, actual measurements should be done on the outside, from call to callback, a delayed callback dispatch can make quite a difference.

michaelnisi:
My motivation for looking into QoS were rare `cpu usage` watchdog terminations.

```
CPU: 144 seconds cpu time over 152 seconds (95% cpu average), exceeding limit of 80% 
```

Apparently caused by Dispatch race conditions:

```
Heaviest stack for the target process:
  12  ??? (libsystem_pthread.dylib + 49428) [0x1f355e114]
  12  ??? (libdispatch.dylib + 74924) [0x1f332f4ac]
  10  ??? (libdispatch.dylib + 44984) [0x1f3327fb8]
  9   ??? (libdispatch.dylib + 124392) [0x1f333b5e8]
  8   ??? (libdispatch.dylib + 124596) [0x1f333b6b4]
  8   ??? (libsystem_kernel.dylib + 149820) [0x1f34dc93c]
```

Of course, I cannot pledge for these coming from Nuke, but since I have set QoS wherever possible throughout my app, I haven’t seen this crash. Mind you, this can be entirely unrelated, mentioned just for context.

Issue #260, opened 2019-06-05
init(memoryCapacity:diskCapacity:diskPath:)' is unavailable in UIKit for Mac


appfrilans:
Using latest Xcode 11.0 beta building for mac I get the following build error:

init(memoryCapacity:diskCapacity:diskPath:)' is unavailable in UIKit for Mac

Would be awesome if that could be fixed! I am using Nuke 7.6.3

kean:
I have Xcode 11 beta but not macOS Catalina yet so I can't use Catalyst. If someone is willing to takes this, PRs are welcome.

appfrilans:
Apparently its not implemented yet, so not sure If we just wait for a fix by Apple or make a workaround.

Apple beta release notes says: "The new required URLCache initializer init(memoryCapacity:diskCapacity:diskPath:) is currently unimplemented. (50739637)" from https://developer.apple.com/documentation/macos_release_notes/macos_10_15_beta_release_notes

bryan1anderson:
This is still not implemented I guess. A workaround would be awesome

bryan1anderson:
@kean @appfrilans What workarounds would be available for this? Has anyone branched to make this work?

kean:
Shipped in [Nuke 8.2.0](https://github.com/kean/Nuke/releases/tag/8.2.0), thanks @ JonathanDowning!

PR #299, opened 2019-09-09
Add Support for Mac Catalyst

JonathanDowning:
This is a proposed resolution to #260

I would point out that the following warning is emitted on Xcode versions lower than 11.0, but this does not prevent compilation:

> Unknown target environment for build configuration 'targetEnvironment'

### DIFF
--- a/Sources/DataLoader.swift
+++ b/Sources/DataLoader.swift
@@ -53,7 +53,7 @@ public final class DataLoader: DataLoading {
         return (200..<300).contains(response.statusCode) ? nil : Error.statusCodeUnacceptable(response.statusCode)
     }
 
-#if !os(macOS)
+#if !os(macOS) && !targetEnvironment(macCatalyst)
     private static let cachePath = "com.github.kean.Nuke.Cache"
 #else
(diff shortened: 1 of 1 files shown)


kean:
Thank you, @JonathanDowning!

kean:
@JonathanDowning, I'm trying to see if it is possible remove the warning. One option that came to mind was to use `#if canImport(AppKit)` instead of `targetEnvironment(macCatalyst)`.

JonathanDowning:
I’ve seen other frameworks work past it by doing `#if canImport(CryptoKit)`, since that’s only available on iOS 13 and later. 
This would silence the warning in Xcode 10 and earlier.
It’s gross, but until Swift has better support for determining iOS versions, I think it could be our best shot!

JonathanDowning:
@kean I cant think of a Swift only solution :/

kean:
Alright, no worries. Let's ship it this way, Xcode 11 is right around the corner already.

JonathanDowning:
OK! Yay :)

PR #305, opened 2019-10-07
Add URLSessionTaskMetrics collection support

markpetrenko:
## Why?

Nuke doesn't have support for URLSessionTaskMetrics collection.

## How?

Added a parameter `sessionTaskMetrics` to `ImageResponse`. Also refactored a bit to make it possible:
Removed tuples from data loading types. Added concrete types:
- DataResponse to represent result of fetching image data
- DataFetchResult enum to represent either chunk response, or complete result of data loading.

### DIFF
--- a/Sources/DataLoader.swift
+++ b/Sources/DataLoader.swift
@@ -11,10 +11,12 @@ public protocol Cancellable: AnyObject {
 public protocol DataLoading {
     /// - parameter didReceiveData: Can be called multiple times if streaming
     /// is supported.
+    /// - parameter didCollectTaskMetrics: Should be called once session finished collecting metrics for the task. See `urlSession(_:task:didFinishCollecting:)` of `URLSessionTaskDelegate` for details.
     /// - parameter completion: Must be called once after all (or none in case
     /// of an error) `didReceiveData` closures have been called.
     func loadData(with request: URLRequest,
                   didReceiveData: @escaping (Data, URLResponse) -> Void,
+                  didCollectTaskMetrics: @escaping (URLSessionTaskMetrics) -> Void,
                   completion: @escaping (Error?) -> Void) -> Cancellable
 }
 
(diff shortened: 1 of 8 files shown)


kean:
Thanks, @markpetrenko! Adding support for `URLSessionTaskMetrics` is definitely something that I would like to add in the framework. I will review the PR as soon as I have some time.

Sources/DataLoader.swift
```diff
@@ -11,10 +11,12 @@ public protocol Cancellable: AnyObject {
 public protocol DataLoading {
     /// - parameter didReceiveData: Can be called multiple times if streaming
     /// is supported.
+    /// - parameter didCollectTaskMetrics: Should be called once session finished collecting metrics for the task. See `urlSession(_:task:didFinishCollecting:)` of `URLSessionTaskDelegate` for details.
     /// - parameter completion: Must be called once after all (or none in case
     /// of an error) `didReceiveData` closures have been called.
     func loadData(with request: URLRequest,
                   didReceiveData: @escaping (Data, URLResponse) -> Void,
+                  didCollectTaskMetrics: @escaping (URLSessionTaskMetrics) -> Void,
```
kean: This is a breaking API change and in the current form will require a major release.

I would suggest considering keeping `DataLoading ` as is and making this a `DataLoader` feature. As as user, for me it would be enough to just be able to log the metrics with their associated URLs. I don't necessarily need them as part of the response (`ImageResponse`).

markpetrenko: Thanks for your feedback! 
In my usage scenario it's more convenient to have metrics in this place.
But you are right, it would be better not to affect the general interface.
I'll push the changes as soon as I can.

markpetrenko: I’ve performed some investigation and think that it would be more useful to have an ability to intercept any URLSessionTaskDelegate call in default implementation of data loader. Clients could log not just URLSessionTaskMetrics, but any moment of network task lifecycle.
what do you think of it? @kean

kean: By intercept do you mean overriding the class and overriding these methods? Yeah, I think it works. The reason why I initially encapsulated everything because I was conservative in terms of how many customization options Nuke provides. The idea was that if the client wants any additional features, they need to roll-out their own DataLoader implementation from scratch. But now, since the framework is mature, I'm ready to provide _all_ of the customization options :)

markpetrenko: Subclassing is a working solution, but I’ve meant something like injection of custom URLSessionTaskDelegate implementation, intended just for proxying original implementation calls. 

But no sure every client needs this. Maybe own DataLoading implementation is still a relevant option for collection URLSessionTaskMetrics. Closing PR for now.

PR #307, opened 2019-10-15
fix demo app label color in dark mode

yurited:
| before | after |
|--------|-------|
| image  | image2  |


kean:
Thank you, @yurited! I completely forgot about Dark Mode in demos. Thanks so much for taking your time and creating this PR.

PR #315, opened 2019-10-28
Added speed manager to check download speed

abdallahnh:
you can check speed by calling SpeedManager.downloadSpeed


kean:
Hi, @abdallahnh! Thanks for putting this PR together.

I'm wondering whether this is something that be shipped as part of the framework. This feature and its implementation seem very specific.

I personally think it would be better to provide APIs to allow users to add features like that on top of the framework instead of building them into the framework itself. For example, `DataLoader` could expose a delegate or a set of closures to be triggered each time a request is started or completed. This could also be useful for other scenarios, for example, for logging.

abdallahnh:
I liked the idea, I'll do that and send you the PR.

PR #316, opened 2019-10-30
Add Closure to know when data received

abdallahnh:
Added didReceiveData to DataLoader, to be used in our projects


kean:
Thanks, @abdallahnh. I'm going to merge this as is. Just to let you know, I might make some minor adjustments later. Adding a way to observe all sorts events which happen in the pipeline is important. I want to see if there is a more general solution which can be used across the framework.

I noticed the CI is now failing because it can't fetch a CocoaPods spec repo. I will investigate also this later.

Thanks, @abdallahnh, for raising this issue and following up with two PRs. Sorry for the long response, I was focusing on other important things, I don't have a lot of time to support the frameworks at the moment.

PR #318, opened 2019-11-06
Fix Void Return Violation

AvdLee:
As returned from the Linter. This simply replaces () with Void where needed.


kean:
Thanks, @AvdLee. This is some recently merged code, I will refactor it before the release.

PR #327, opened 2019-11-12
RoundedCorners and Circle image processor Border width

ejensen:
The `RoundedCorners` image processor has an existing `Border` parameter. However It was not respecting the `border.width` when drawing the stroke. The stoke was always drawn with a width of 1.
This PR corrects the width of the stroke border in the `RoundedCorners` processor. 

This PR also adds an optional `border` parameter to the `Circle` processor. The new parameter defaults to `nil` in order to prevent a breaking API change.

### DIFF
--- a/Sources/ImageProcessing.swift
+++ b/Sources/ImageProcessing.swift
@@ -157,14 +157,22 @@ extension ImageProcessor {
 
     /// Rounds the corners of an image into a circle.
     public struct Circle: ImageProcessing, Hashable, CustomStringConvertible {
-        public init() {}
+        private let border: Border?
+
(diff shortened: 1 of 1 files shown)


kean:
Thanks, @ejensen! This looks good. I don't remember why `Border` was implemented this way initially, it's great that it now works as expected.

I'm going to merge this and I will also add a `typealias ImageProcessor. RoundedCorners.Border = ImageProcessor.Border` to make sure there are no breaking API changes. Thanks again. If you see any other issues in processors, please let me know. They are a new addition in Nuke 8, not as polished yet.

PR #338, opened 2020-03-01
Border width Scale of RoundedCorners image processor 

kazumanagano:
ImageProcessing.Border has width parameter. However, now border width is not drawn as expected, because, border width does not take `Screen.scale & image.scale` into account.

This PR corrects the width caluculation of the stroke border in the RoundedCorners processor.


kean:
Hi, @kazumanagano. Thanks for the patch. 

I understand why we might want to take `Screen.scale` into account. I'm not sure about `image.scale`. Could you please clarify what issues it solves and how?

kean:
I think I now what the issue is. To make sure that the borders look the way you expect, you must make sure that the size of the image exactly matches the size of the view in which it gets displayed.

Issue #364, opened 2020-05-04
[TSan] Running Unit tests with Xcode 11.4.1 causes data race


LucianoPAlmeida:
Hey :)
Running Unit tests on XCode 11.4.1 makes TSan not happy
I seen that CI runs only on XCode 11.3 ... but a 11.4 is already available here https://changelog.travis-ci.com/xcode-11-4-released-144748

![Screen Shot 2020-05-04 at 16 02 26](https://user-images.githubusercontent.com/8292651/81003355-24c24d80-8e21-11ea-95e2-5b337b56c90f.png)

kean:
Thanks for reporting. TSan, my worst nightmare. I was not able to repro so far on Xcode 11.4.1 running locally. Can you share the steps? Are you able to repro every time?

LucianoPAlmeida:
Ahh sure, I also enabled TSan for Test action on Nuke Unit Tests Target :) 
So may not be an XCode 11.4.1 issue, seems just not caught because is not enable on Unit Tests ...

kean:
Fixed by https://github.com/kean/Nuke/issues/364

PR #365, opened 2020-05-04
Fixing Nuke.Operation TSan issues with non atomic properties.

LucianoPAlmeida:
This is a simple fix to #364


kean:
Thanks, I think this should do it. I tried to avoid creating unnecessary locks, but I would rather keep TSan happy. I don't want users to see any issues in their usage.

LucianoPAlmeida:
> Thanks, I think this should do it. I tried to avoid creating unnecessary locks, but I would rather keep TSan happy. I don't want users to see any issues in their usage.

That is a good thing since unnecessary locks can get in the way of performance-critical code
Anyways, thanks for the review :)

kean:
I'm not using these operation on the main queue, so it should be good.

PR #370, opened 2020-05-22
Use DispatchSourceMemoryPressure

docterd:
Use DispatchSourceMemoryPressure instead of UIApplication.didReceiveMemoryWarningNotification. This should work in other targets then iOS and tvOS as well.

Unfortunately I'm not sure how to cover this in a test. Maybe it would be better to use it only on other targets and keep the current implementation on iOS/tvOS


kean:
Thanks, @docterd. It's a great improvement, especially for watchOS. It's fine not to cover this with unit tests, it's straightforward.

PR #371, opened 2020-05-23
Add tint colors to ImageLoadingOptions

basememara:
When retrieving icons from the cloud, it's useful to set tint colours to them based on the app theme. I've made the assumption if tint colours are set, then rendering mode must be set to `.alwaysTemplate` which I think is a fair assumption? Looking forward to hearing your thoughts or if there's a better place to implement this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ func tableView(_ tableView: UITableView, cellForItemAt indexPath: IndexPaths) ->
 
 > When the view is deallocated, an associated request gets canceled automatically. To manually cancel the request, call `Nuke.cancelRequest(for: imageView)`.
 
-### Placeholders, Transitions, Content Modes
+### Placeholders, Transitions, Content Modes, Tint Colors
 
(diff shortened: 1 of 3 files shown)


kean:
Thanks, @basememara. I like the approach where the rendering mode is automatically determined based on the `tintColor`. I think this looks good.

These is a little SwiftLint warning on the CI. No worries, I can take care of it in `master`.

P.S. It seems there might be a need to generalize/redesign how some options are set per mode (success/failure). I don't think it should be done in the scope of this PR, I'm just thinking out loud. It kind of reminds me of states in `UIButton`, so maybe somewhat similar would work.

basememara:
Thanks for the merge and taking care of the lint. Awesome library btw, very clean.

I see what you're saying about the states. Just some thoughts but maybe some kind of extension like below, even if it was just internal:

```swift
image.setContentMode(.scaleAspectFit, for: .downloaded)
image.setContentMode(.scaleAspectFit, for: .placeholder)
image.setTintColor(.red, for: .failure)
```

PR #372, opened 2020-05-24
Fixed a little misspells in documentation

mar-comp:


kean:
Thanks for the fixes, I appreciate it!

PR #374, opened 2020-06-02
Fix RateLimiter

TangentW:


kean:
Thanks for the fix! Without it, the delay is currently always at least 100 ms instead of the intended 14 ms. It seems to be working reasonably well. I should probably revisit the default configuration, 14 ms seems to be a bit aggressive.

PR #377, opened 2020-06-15
Fix a bug where ImagePipeline.loadImage could retain its completion block indefinitely

ken-broadsheet:
A task can be synchronously terminated by its `starter` block during a call to Task.subscribe(), for instance when the result is cached. This in turn causes ImagePipeline's startImageTask/startDataTask to retain the task and subscription in the `tasks` dictionary indefinitely.

This fix changes Task.subscribe() to return nil if the task is terminated before the method returns. This appears to be in line with the documented intent for the method, and resolves the issue.

### DIFF
--- a/Sources/ImagePipeline.swift
+++ b/Sources/ImagePipeline.swift
@@ -887,3 +887,10 @@ public extension ImagePipeline {
         }
     }
 }
+
+// MARK: - Testing
+internal extension ImagePipeline {
+    var taskCount: Int {
+        return tasks.count
+    }
+}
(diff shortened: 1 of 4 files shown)


kean:
Oh wow, that’s a nice catch. I wonder how you discovered this issue.

I’m going to merge this as soon as I’m able to review on my computer, thank you!

ken-broadsheet:
Thank you! I was trying to use loadImage to prefetch, and noticed that an object referenced by the completion block was sticking around longer than it should have.

Issue #378, opened 2020-06-22
Cannot build on Nuke in Xcode 11.5


dogehk:
Hi, all

I checked the error that is "Type 'String' has no member 'priorityRequestLow'" in line number 294 on ImageProcessing.swift.

Xcode Version 11.5
Swift version 5.2.4

Thanks, Regards.

PR #379, opened 2020-06-22
Supported the 11.5 of Xcode.

dogehk:
Related to issue #378

### DIFF
--- a/Sources/ImageProcessing.swift
+++ b/Sources/ImageProcessing.swift
@@ -291,7 +291,7 @@ extension ImageProcessors {
 
         /// A default context shared between all Core Image filters. The context
         /// has `.priorityRequestLow` option set to `true`.
-        public static var context = CIContext(options: [.priorityRequestLow: true])
+        public static var context = CIContext(options: [kCIContextPriorityRequestLow : true]);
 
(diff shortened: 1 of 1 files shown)


kean:
Hi @dogehk. I'm also using Xcode 11.5 for development, and I tested Nuke on it. I don't see these issues with my setup. I'm wondering what the root cause is. We are using an initializer that works with [CIContextOption](https://developer.apple.com/documentation/coreimage/cicontextoption).

saqibomer:
Issue persists in Xcode 12.4. 
9.0 and 10.0 have 
`public static var context = CIContext(options: [.priorityRequestLow: true])`
Instead of 
`public static var context = CIContext(options: [kCIContextPriorityRequestLow: true])`

kean:
Hi, `kCIContextPriorityRequestLow` gets imported as [`.priorityRequestLow`](https://developer.apple.com/documentation/coreimage/cicontextoption/1620424-priorityrequestlow) in Swift. It's been like that for a couple of years now.

Could you describe how you compiling it?

saqibomer:
It was due to build configuration. I had set older swift version in pod configuration.

PR #383, opened 2020-06-23
Make imageView unowned in ImageViewController

ismetanin:
Hi! Looks like the referenced bug was fixed long ago. So, here it is - `unowned imageView`
I'm not sure if it changes any performance characteristics but theoretically, it should, right?

### DIFF
--- a/Sources/ImageViewExtensions.swift
+++ b/Sources/ImageViewExtensions.swift
@@ -332,9 +332,7 @@ public struct ImageLoadingOptions {
 /// however as it stands today `ImageViewController` is just a helper class,
 /// making it public wouldn't expose any additional functionality to the users.
 private final class ImageViewController {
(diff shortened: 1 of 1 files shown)


kean:
Thanks, let me test it, I'm interested to see whether it is going to affect the performance in a significant way. But regardless of the results, I would probably prefer to leave `weak` there and remove the comment instead. `unowned` is typically used for situations where you can guarantee that the object is not going to be deallocated while the reference is in use. That's not the case here.

ismetanin:
@kean Alright, should I change on `weak` now or we'll wait for your test results?

kean:
I would keep it as `weak`. I'm curios to test it, but I don't expect this change to be measurable.

ismetanin:
@kean okay, I changed to weak. Anyway, if you find something valuable in measurements share it somewhere please I was always interested to find any case where `unowned` really beats `weak`

PR #384, opened 2020-06-24
Support XCFramework

kean:
- [x] Update project to support binary distribution
- [ ] Update package description to support binary distribution (was not able to support both binary and source distribution at the same time)
- [x] Automate the creation of binary frameworks

> Info https://developer.apple.com/videos/play/wwdc2020/10147/


kean:
I'm just testing it for now. I don't think there is much sense in supporting binary distribution moving forward. If you support all Apple platforms, the archive becomes massive. For my use case, it's the client that should be able to easily pre-compile packages.

PR #386, opened 2020-07-24
Fix broken link in Nuke 9 Migration Guide

rlwimi:
Looks like the link to release notes is broken for pointing at `9.0-beta.2` instead of `9.0.0-beta.2`. Anyways, it seems like this should be pointing at `9.0.0` now.


kean:
Thanks!

PR #392, opened 2020-08-25
Fix cache memory pressure monitoring

ejensen:
Fixes an issue introduced in #370 where caches are cleared when the memory pressure changes to a `normal` level.
Limits image cache clearing to `warning` and `critical` memory pressure level changes.


kean:
Thanks, @ejensen. That's great, thanks for spotting this, I missed it during the review.

kean:
I wonder what's the difference between `warning` and `critical` and whether the handling should be different for these two scenarios. E.g. for `warning` we could remove just some images - similar to what `ImageCache` already does when apps enters background.

Issue #389, opened 2020-08-12
416 range error


mtrezza:
### Issue

I see a significant amount of `416 Range Not Satisfiable` logs with our CDN. These errors only occur in the iOS app with Nuke, not in the Android app with Glide, which is loading the same images. However, I have not seen any issues with image loading in the app and there are no user complaints about images not loading.

### Logs

Status | Timestamp | Remote IP | Size | Path
-- | -- | -- | -- | --
416 | 2020-07-31 01:55:42 | (REDACTED) | 598 Bytes | (REDACTED)

The "size" in the logs which I assume is the requested range is always around 400 - 600 Bytes for these 416 errors.
For successful CDN responses, "size" is the actual image size, like several KB.

### Question

Are there any cases in which Nuke uses range requests?

kean:
Hi, yes, by default Nuke uses [range requests](https://github.com/kean/Nuke/blob/master/Sources/Internal.swift#L255) and this error code is most likely related to them. To disable it, you can set `isResumableDataEnabled` to `false` on the pipeline configuration. However, it would be great to investigate and try to fix it.

kean:
Nuke has a very [simple algorithm](https://github.com/kean/Nuke/blob/master/Sources/Internal.swift#L298) for setting requested ranges, I'm not sure what can be wrong with the range.

cjmconie:
Hi, I am experiencing a similar issue. 

It seems to appear when there are concurrent requests, for example inside a `UICollectionView` displaying a grid of `UICollectionViewCell`s . I'm using `Nuke.loadImage(with:into:)`. `ImagePipeline` is unmodified. Images are being requested directly from Firebase Storage. The error from Nuke is: `Failed to load image data: Response status code was unacceptable: 416`

Customising the `ImagePipeline` by setting `isResumableDataEnabled` as `false` resolves the issue.

Is there any more information I can provide to help?

mattczech:
Hello, I'm also seeing occasional 416 errors.

In my case, I'm loading images stored in S3. Each 416 that I've observed is a result of Nuke requesting a starting byte count that is equal to the size of the image.

I'm using a `UICollectionView` to display a grid of images with `Nuke.loadImage(with:into:)`. No modifications to the pipeline.

When I've seen the error, I've observed the following:

- The image exists in the URLCache and it's not expired.
- The collection view's initial layout occurs, cells are configured and images start loading. During this, one of the cells starts loading a different image.
[...]

PR #402, opened 2020-11-16
#389 Fix an issue where fully downloaded data will sometimes be saved as resumable

kean:
Fixes https://github.com/kean/Nuke/issues/389

### DIFF
--- a/Sources/Internal.swift
+++ b/Sources/Internal.swift
@@ -260,6 +260,7 @@ struct ResumableData {
         // Check if "Accept-Ranges" is present and the response is valid.
         guard !data.isEmpty,
             let response = response as? HTTPURLResponse,
+            data.count < response.expectedContentLength,
             response.statusCode == 200 /* OK */ || response.statusCode == 206, /* Partial Content */
             let acceptRanges = response.allHeaderFields["Accept-Ranges"] as? String,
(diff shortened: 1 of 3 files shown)


Sources/Internal.swift
```diff
@@ -260,6 +260,7 @@ struct ResumableData {
         // Check if "Accept-Ranges" is present and the response is valid.
         guard !data.isEmpty,
             let response = response as? HTTPURLResponse,
+            data.count < response.expectedContentLength,
```
kean, line 263: URLSession always seem to return cached data in a single chunk, so this should always prevent cached data from being saved as resumable data. Ideally, I would like to explicitly check if data is coming from cache, but this doesn't seem to be possible.

kean: This should also cover the following (rate) scenario:

- Data fully downloaded form the server
- `func imageDataLoadingTask(_ task: OriginalImageDataTask, context: OriginalImageDataTaskContext, didReceiveData chunk: Data, response: URLResponse, log: Log)` is called with the last chunk
- The request is cancelled and resumable data is saved
- `func imageDataLoadingTask(_ task: OriginalImageDataTask, context: OriginalImageDataTaskContext, didFinishLoadingDataWithError error: Swift.Error?)` is called

PR #427, opened 2021-01-11
Use image scale instead of screen

arnopoulos:
This PR changes how points are calculated. Currently we use the screen's scale factor to calculate the image's point scale but this only works if the image's scale matches the screen's scale. This PR fixes this issue by resizing based off of image scale rather than screen scale to account for when the image scale is not equal to the screen's scale.


kean:
Hey, @arnopoulos. Thanks for opening the PR. Could you please describe what is the use case? The current resizer behavior is pretty simple. You either the specify the target size in pixels or points. It ignores the image scale by design. What are the scenarios in which you need to take image scale into account?

By the way, talking about image scale in general. By default, it's currently set to match the screen scale. The users can customize it if needed.

kean:
Hey, @arnopoulos. I'm going to close this since there was not response from your side for over a month. Please feel free to refresh this PR if this is still something you are interested in adding. I'm open to suggestions, there appears to be a room for improvement, but I'm just not fully up to speed with your suggestion.

PR #439, opened 2021-02-12
macOS: Remove requirement of NSView 

imthath-m:
At the moment, for working with Nuke in macOS, all views need to be a subclass of `NSView`. This is because the `ImageDisplayingView` enforces that constraint, which however is not needed for Caching purposes (the main reason I'm using Nuke). 

Hence, I have removed the `NSView` requirement and used `NSObject` instead, so that it works with `NSCell`, `NSTextAttachment` and other types which can display images but do not inherit from `NSView`. To make the animation work in layer based views, I have added a `Nuke_Animating` protocol only for macOS.

### DIFF
--- a/Sources/ImageViewExtensions.swift
+++ b/Sources/ImageViewExtensions.swift
@@ -42,8 +42,18 @@ extension UIImageView: Nuke_ImageDisplaying {
 }
 #elseif os(macOS)
 import Cocoa
-/// An `NSView` that implements `ImageDisplaying` protocol.
-public typealias ImageDisplayingView = NSView & Nuke_ImageDisplaying
+/// An `NSObject` that implements `ImageDisplaying`  and `Animating` protocols.
+/// Can support `NSView` and `NSCell`. The latter can return nil for layer.
+public typealias ImageDisplayingView = NSObject & Nuke_ImageDisplaying & Nuke_Animating
+
+@objc public protocol Nuke_Animating {
+
+  @objc var layer: CALayer? { get }
+}
+
+// A view already has a layer property definied in AppKit.
+// Hence, no action required for conforming to Animating protocol.
+extension NSView: Nuke_Animating { }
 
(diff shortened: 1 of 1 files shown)


kean:
Thanks, it's getting a bit late here, going to review some time tomorrow.

kean:
This looks good, thanks. I'll might make some changes to the way animations are handled later.

Are you planning to add `Nuke_ImageDisplaying` conformance to `NSCell`?

imthath-m:
I initially started off in that path. But my requirement at the moment can be fulfilled by conforming `NSTextAttachment` to `Nuke_ImageDisplaying`. Since we are using other subclasses of `NSCell` also in our project in multiple places, we might need that also in the future.

imthath-m:
Thanks for the quick response, @kean. I hope this goes out with 9.3.0. When are planning to release it?

kean:
I'm going to push an update soon, 9.3.0 is in beta https://github.com/kean/Nuke/releases/tag/9.3.0-beta.1. If you could try it (or install from master) and confirm there are no regressions that would expedite the process. It has major internal changes to improve pipeline performance.

kean:
FYI I slightly reworked the solution. `layer` added to `Nuke_ImageDisplaying` on macOS, added default implementation for `NSView`. This should be fully backward compatible.

imthath-m:
That will work perfect. Thanks for that. 

Before we could decide on when to take up the beta and test, you have shipped the public version. We couldn't match your speed.

Sloaix:
There may be a problem with this merge, has it been tested ?

I rolled back to the 9.3.0-beta.1, previous version of 9.3.0 , everything is fine.😂

my development environment is 

```
Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
```

WX20210623-162602
WX20210623-162550
WX20210623-163015

PR #489, opened 2021-06-16
Fix: doc comment

ryu1sazae:
Thank you for the nice update to version10!!
This is just a small fix of doc comment.


kean:
Thanks!

PR #490, opened 2021-06-16
Add: default parameter nil to nuke_display(image:data:)

ryu1sazae:
Thank you for the nice update to version10!!

I wanna suggest one change. nuke_display(image:data) which is a protocol method of Nuke_ImageDisplaying is defined as extension method of UIImageView, NSImageView and WKInterfaceImage. But, the data parameter is never used although this can be the destructive change at updating to version10.
So, if there are not some hesitations for this point, it is better to have default parameter nil, I think.


kean:
Thanks, yeah it might ease the migration a bit.

PR #494, opened 2021-07-05
Improved runtime safety of ImagePublisher

Nickersoft:
This PR moves around the logic of the Nuke ImagePublisher to only emit when demand is requested. Emitting otherwise can potentially lead to unexpected crashing, specifically in the case of returning caching images. The previous logic would emit a cached value immediately after sending the subscription to subscribers, and would sometimes crash if the value was sent before the subscribers could fully subscribe.

The approach used here was influenced by articles such as [this](https://medium.com/flawless-app-stories/swift-combine-custom-publisher-6d1cc3dc248f) and [this](https://www.swiftbysundell.com/articles/building-custom-combine-publishers-in-swift/#keeping-up-with-demand).

### DIFF
--- a/Sources/Combine/ImagePublisher.swift
+++ b/Sources/Combine/ImagePublisher.swift
@@ -38,19 +38,49 @@ public struct ImagePublisher: Publisher {
     }
 
     public func receive<S>(subscriber: S) where S: Subscriber, Failure == S.Failure, Output == S.Input {
-        let subscription = ImageSubscription()
-        subscriber.receive(subscription: subscription)
+        let subscription = ImageSubscription(
+            request: self.request,
+            pipeline: self.pipeline,
+            subscriber: subscriber
+        )
 
+        subscriber.receive(subscription: subscription)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


kean:
Thanks, @Nickersoft. I appreciate you looking into this and coming up with a fix! I had a note to change it, but I didn't expect the initial implementation to be this problematic.

The MR looks good, but we'll probably need to handle a scenario where `request()` is called more than once.

Nickersoft:
@kean Thanks for the merge! And yeah, I think it was only problematic maybe under a certain set of super specific conditions that my program just happened to meet haha. I did some logging in request() and it looked like it was only called once per subscription request. I think it is a method called per subscription to inform the publisher of what the expected demand will look like.. can't find any docs confirming this, but it seems to be [the implication](https://www.raywenderlich.com/books/combine-asynchronous-programming-with-swift/v2.0/chapters/2-publishers-subscribers#toc-chapter-005-anchor-005). So I'm not sure we'd have to worry about request() being called multiple times unless someone did so directly. As for thread safety, AFAIK the implementation is thread safe, but if there's ever an issue we could add some kind of lock for file reads.

PR #506, opened 2021-09-02
Migrate video decoding from NukeUI

larryonoff:


Sources/Core/Video/VideoDecoding.swift
```diff
@@ -0,0 +1,85 @@
...
+            guard let preview = makePreview(for: data) else {
+                return nil
+            }
+            didProducePreview = true
+            return ImageContainer(image: preview, type: .mp4, isPreview: true, data: data)
+        }
+
+        private static var isRegistered: Bool = false
+
+        public static func register() {
```
kean, line 52: This decoder can be registered by default in `ImageDecoderRegistry` which NukeUI didn't have access to.

Sources/Core/Video/VideoDecoding.swift
```diff
@@ -0,0 +1,85 @@
...
+    return PlatformImage(cgImage: cgImage)
+    #endif
+}
+
+// TODO: extened support for other image formats
+// ftypisom - ISO Base Media file (MPEG-4) v1
+// There are a bunch of other ways to create MP4
+// https://www.garykessler.net/library/file_sigs.html
+private func match(_ data: Data, offset: Int = 0, _ numbers: [UInt8]) -> Bool {
+    guard data.count >= numbers.count + offset else { return false }
```
kean, line 79: There is an existing `_match()` method for that in `ImageType` which should be reused. Also, should the `ImageType` be extended? I think yet, but it can be awkward to have video there. But `GIF` is basically video too.

Sources/Core/Video/AVAsset.swift
```diff
@@ -0,0 +1,75 @@
...
+// Copyright (c) 2015-2021 Alexander Grebenyuk (github.com/kean).
+
+import AVKit
+import Foundation
+
+#if !os(watchOS)
+
+extension ImageContainer {
+    /// A fetched video from data.
+    public var asset: AVAsset? {
```
kean, line 12: Should it be created once on `ImageContainer` initialization?

Sources/Core/Video/AVAsset.swift
```diff
@@ -0,0 +1,75 @@
...
+        guard type == .mp4, let data = data else {
+            return nil
+        }
+        return AVDataAsset(data: data)
+    }
+}
+
+// This allows LazyImage to play video from memory.
+final class AVDataAsset: AVURLAsset {
+    private let _loader: DataAssetResourceLoader
```
kean, line 22: I generally don't use underscores. I think SwiftLint won't like that when I run the CI.

Sources/Core/Video/VideoDecoding.swift
```diff
@@ -0,0 +1,85 @@
...
+extension ImageDecoders {
+    public final class Video: ImageDecoding, ImageDecoderRegistering {
+        private var didProducePreview = false
+
+        static func isVideo(_ data: Data) -> Bool {
+            match(data, offset: 4, [0x66, 0x74, 0x79, 0x70])
+        }
+
+        public var isAsynchronous: Bool {
+            false
```
kean, line 24: I think it should be asynchronous depending on how slow the creation of an asset is (assuming it will be moved here). Async means "executed on an operation queue". An alternative is executing on the image pipeline internal serial queue (slightly less overhead for quick operations).

Sources/Core/Video/VideoDecoding.swift
```diff
@@ -0,0 +1,85 @@
...
+        public init?(partiallyDownloadedData data: Data, context: ImageDecodingContext) {
+            guard Video.isVideo(data) else { return nil }
+        }
+
+        public func decode(_ data: Data) -> ImageContainer? {
+            ImageContainer(image: PlatformImage(), type: .mp4, data: data)
+        }
+
+        public func decodePartiallyDownloadedData(_ data: Data) -> ImageContainer? {
+            guard !didProducePreview else {
```
kean, line 40: This will definitely require some unit tests.

Sources/Core/Video/VideoDecoding.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+private func makePreview(for data: Data) -> PlatformImage? {
+    let asset = AVDataAsset(data: data)
+    let generator = AVAssetImageGenerator(asset: asset)
+    guard let cgImage = try? generator.copyCGImage(at: CMTime(value: 0, timescale: 1), actualTime: nil) else {
+        return nil
+    }
+    #if os(macOS)
+    return PlatformImage(cgImage: cgImage, size: .zero)
```
kean, line 68: There are platform compatibility extensions in Nuke. You can simply use `PlatformImage(cgImage)`.

kean (approved):
Hey, thanks for putting this together. I left some comments. If you want to dig into it, please feel free to do so. Otherwise I can merge it, and address the comments myself later.

larryonoff:
@kean I've resolved your comments, except one. Please check.

kean:
Merging. There are some build issue. I'm going to fix them in master.

kean:
Interesting, checking `0x66, 0x74, 0x79, 0x70` in the signature wasn't enough. Apple heic files have the same signature!

PR #515, opened 2021-10-24
Async/Await

kean:
### 1. Add `ImagePipeline` `loadImage` async method

```swift
do {
    return try await pipeline.loadImage(with: request)
} catch {
    guard let error = (error as? ImagePipeline.Error),
          (error.dataLoadingError as? URLError)?.networkUnavailableReason == .constrained else {
        throw error
    }
    return try await pipeline.loadImage(with: lowQualityImageURL)
}
```

### 2. `FetchImage` now supports async responses

```swift
struct ImageView: View {
    let url: URL

    @StateObject private var image = FetchImage()

    var body: some View {
        ZStack {
            Rectangle().fill(Color.gray)
            image.view?
                .resizable()
                .aspectRatio(contentMode: .fill)
                .clipped()
        }
        .onAppear { loadImage(url) }
        .onChange(of: url) { loadImage($0) }
        .onDisappear(perform: image.reset)
    }

    private func loadImage(url: URL) {
         image.load {
               try await ImagePipeline.shared.loadImage(with: url) // Or could be any async task you want
         }
    }
}
```

### DIFF
--- a/Sources/Core/ImagePipeline.swift
+++ b/Sources/Core/ImagePipeline.swift
@@ -148,13 +148,17 @@ public final class ImagePipeline {
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     public func loadImage(with request: ImageRequestConvertible) async throws -> ImageResponse {
         final class TaskBox {
-            var task: ImageTask? // Ideally, ImageTask should be Sendable
+            var task: ImageTask?
         }
         let box = TaskBox()
         return try await withTaskCancellationHandler(handler: {
             box.task?.cancel()
         }, operation: {
             try await withUnsafeThrowingContinuation { continuation in
+                // The pipeline guarantees that the callbacks (either onCancel or
+                // completion) are called exactly once. `onCancel` is a new addition
+                // just for Async/Await. Ideally, the completion should be called on
+                // cancellation instead, but that ship has sailed. Maybe in Nuke 11.
                 box.task = loadImage(with: request.asImageRequest(), isConfined: false, queue: nil, progress: nil, onCancel: {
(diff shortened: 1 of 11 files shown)


Sources/Core/ImagePipeline.swift
```diff
@@ -138,6 +138,33 @@ public final class ImagePipeline {
...
+        let box = TaskBox()
+        return try await withTaskCancellationHandler(handler: {
+            // TODO: How can we report cancellation from here? Add cancellation to ImageTask?
+            box.task?.cancel()
+        }, operation: {
+            try await withUnsafeThrowingContinuation { continuation in
+                box.task = self.loadImage(with: request, queue: nil, progress: nil) {
+                    continuation.resume(with: $0)
+                }
+            }
```
sindresorhus, line 164: Ideally, you would make `ImageTask` conform to `Sendable` and then use it like this:

```suggestion
        var task: ImageTask?

        return try await withTaskCancellationHandler(handler: { [task] in
            // TODO: How can we report cancellation from here? Add cancellation to ImageTask?
            task?.cancel()
        }, operation: {
            try await withUnsafeThrowingContinuation { continuation in
                task = self.loadImage(with: request, queue: nil, progress: nil) {
                    continuation.resume(with: $0)
                }
            }
```

kean: That's a good idea, thanks! Initially, I just went with what the compiler suggested. I'm planning to wrap this up today.

kean: It doesn't seem that I'll be able to make `ImageTask` `Sendable`. It's not thread-safe and has to be thread-confined. Cancellation is.

PR #523, opened 2021-12-21
Support m4v

kor45cw:
I tried to play m4v through NukeUI, but found an issue where the conversion failed in ImageContainer. Post a PR that fixes the problem.


kean:
Thanks!

Issue #533, opened 2022-02-01
Support for .mov / Alpha Channel for Videos with Transparency


SENTINELITE:
Hey! 

I was using SDWebImage, but have been wanting to switch over to Nuke, for my project.

Unfortunately, there's no support for videos with an alpha channel. I have transparent gifs/apngs, but as you know, they aren't performant. 

As a temporary hack, I'll  just render out a flat mp4, & apply some transparency to the video.

Ideally, I'd like to help contribute & push to support .mov/proRes etc. Just don't want to start, if it's too far out of scope here.

kean:
Hi,

Nuke and [NukeUI](https://github.com/kean/NukeUI) support (short) videos out of the box, but it's still an experimental feature, and it doesn't support all formats yet.

For decoding, see [ImageDecoders+Video](https://github.com/kean/Nuke/blob/master/Sources/Core/Decoding/ImageDecoders%2BVideo.swift). Rendering is done using [AVDataAsset](https://github.com/kean/Nuke/blob/master/Sources/Internal/AVDataAsset.swift) (in-memory video representation) and views from [NukeUI](https://github.com/kean/NukeUI).

The project is open for PRs in this area. The [demo project](https://github.com/kean/NukeDemo) has an example of basic video playback.

SENTINELITE:
I submitted a PR to add initial support for .mov files. It's successfully working on my end. The video format is ProRes 4444. I've attached a video, for testing purposes, as well. #534

https://user-images.githubusercontent.com/44782976/152077725-dd1b9e5d-85a3-402a-921a-5cde9f9f6b68.mov

PR #534, opened 2022-02-02
Add support for .mov files.

SENTINELITE:
This is the initial version for adding .mov file support. Issue #533


kean:
Hey, thanks for taking your time to work on it. The code looks good. Yes, I think the only missing part was decoding, and now that it's covered everything else should also work.

What about the TODO? I think line 28 looks ok as is.

SENTINELITE:
I wasn't sure if we needed to coalesce some sort of .mov type? I haven't really dealt much with stuff related on this line, so I wasn't sure. 😅

Issue #544, opened 2022-05-24
Error.decodingFailed is missing corrupted `Data`


freak4pc:
Heya 👋 
We noticed in our logs a small wave of sporadic `ImagePipeline.Error.decodingFailed` error, but we don't have a way to debug those / add logs since the error doesn't include any additional information.

From what I checked it can fail for an empty Data, empty Response or an invalid image data. 
We'd like to know which one It is - but also, if it's the latter, know exactly what is the data that caused it to fail decoding.

Would you accept a PR that adds the `Data` as an associated value to the error? Or any other idea that comes to mind?

Thanks !

freak4pc:
> A common reason for this is a misbehaving API, e.g. returning HTML instead of image data.
Yes, I definitely suspect this :) But without the data it's hard to know exactly what happened.

I'll put a PR soon. Thanks!

kean:
Thanks for the PR. I'm planning to release it in the scope of Nuke 11*  with other major improvements to error handling. I want each error to have as much context as possible to localize a problem. For example, here is a [first related PR](https://github.com/kean/Nuke/pull/551) that makes  `ImageProcessing` throwing.

*It's technically a breaking change. Maybe if data had a default value, it could be source-compatible. But I'm not sure it's worth releasing it as is. I suggest installing from the main branch for now.

freak4pc:
Is there a plan for v11? 
We have a nasty bug we want to investigate and have zero visibility about why decoding fails.

kean:
[Nuke 11 (Beta 1)](https://github.com/kean/Nuke/releases/tag/11.0.0-beta.1) and I'm making error reporting much-much better, among many other changes.

PR #545, opened 2022-05-25
Add associated Data to ImagePipeline.Error.decodingFailed

freak4pc:
Based on discussion at #544

### DIFF
--- a/Sources/Core/Tasks/TaskFetchDecodedImage.swift
+++ b/Sources/Core/Tasks/TaskFetchDecodedImage.swift
@@ -31,14 +31,14 @@ final class TaskFetchDecodedImage: ImagePipelineTask<ImageResponse> {
         // Sanity check
         guard !data.isEmpty else {
             if isCompleted {
-                send(error: .decodingFailed)
+                send(error: .decodingFailed(data))
             }
             return
         }
 
         guard let decoder = decoder(data: data, urlResponse: urlResponse, isCompleted: isCompleted) else {
             if isCompleted {
-                send(error: .decodingFailed)
+                send(error: .decodingFailed(data))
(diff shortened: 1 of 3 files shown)


kean:
Could you please update the unit tests? If you run them locally, you'll see what the issues are.

freak4pc:
@kean my bad, I opened Package.swift directly and it doesn't have the tests targets :)
Fixed.

Sources/Core/Tasks/TaskFetchDecodedImage.swift
```diff
@@ -68,7 +68,7 @@ final class TaskFetchDecodedImage: ImagePipelineTask<ImageResponse> {
         if let response = response {
             send(value: response, isCompleted: isCompleted)
         } else if isCompleted {
-            send(error: .decodingFailed)
+            send(error: .decodingFailed(response?.container.data ?? Data()))
```
kean, line 71: @freak4pc, I was reworking this code for Nuke 11 and just noticed that it never attaches data. In this code block `response` is always `nil`.

freak4pc: That's interesting. So where would we be able fetching the "right" Data from?

kean: I'll fix it in the patch release later today.

kean: The issue is that `ImageResponse` is always `nil` in this block – decoding failed.

kean: Fixed in [10.11.1](https://github.com/kean/Nuke/releases/tag/10.11.1).

PR #560, opened 2022-06-14
Draft: DataLoading protocol now uses Async/Await

kean:


kean:
It's probably best to leave it for Nuke 12 when I'm planning to rewrite all of the internals using Structured Concurrency.

PR #571, opened 2022-06-28
ImageTaskDelegate Concurrency Problem [Feedback Needed]

kean:
## Problem

The new async/await [`image(for:delegate:)`](https://kean-docs.github.io/nuke/documentation/nuke/imagepipeline/image(for:delegate:)) method in unreleased [Nuke 11.0](https://github.com/kean/Nuke/pull/547) has the following signature:

```swift
func image(
    for request: any ImageRequestConvertible,
    delegate: ImageTaskDelegate? = nil
) async throws -> ImageResponse
```

[`ImageTaskDelegate`](https://kean-docs.github.io/nuke/documentation/nuke/imagetaskdelegate) sends task-level events: task created, progress updated, progressive preview produced, etc. The design is inspired by the [async/await API](https://developer.apple.com/documentation/foundation/urlsession/3767352-data) added to `URLSession` in iOS 15.

```swift
public protocol ImageTaskDelegate : AnyObject, Sendable {
    func imageTaskCreated(ImageTask)
    func imageTask(ImageTask, didReceivePreview: ImageResponse)
    func imageTask(ImageTask, didUpdateProgress: ImageTask.Progress)
    // And more...
```

The delegate is captured weakly and its methods are called on the main queue by default, but can be changed using [`callbackQueue`](https://kean-docs.github.io/nuke/documentation/nuke/imagepipeline/configuration-swift.struct/callbackqueue). The problem is that this design doesn't work well with structured concurrency.

**Example**

The delegate is most likely going to be used on the main queue most of the time, e.g. in custom views which are typically @MainActors.

```swift
@MainActor
private final class AsyncImageView: UIView, ImageTaskDelegate {
    private var imageTask: ImageTask?
    private let imageView = _ImageView()

    deinit {
        imageTask?.cancel()
    }

    func loadImage() async throws {
        imageView.image = try await pipeline.image(for: url, delegate: self).image
    }

    // This gets called immediately from `image(for:)` (important! invariant)
    func imageTaskCreated(_ task: ImageTask) {
        self.imageTask = task
    }

    func imageTask(_ task: ImageTask, didReceivePreview response: ImageResponse) {
        imageView.image = response.image // Display progressively decoded preview
    }
}
```

Unfortunately, this code produces the following warning:

Screen Shot 2022-06-27 at 8 15 45 PM

### Option 1

Make [`ImageTaskDelegate`](https://kean-docs.github.io/nuke/documentation/nuke/imagetaskdelegate) `@MainActor`. It'll require a few changes to the pipeline, but it's feasible.

Cons:

- The user can no longer configure the callback queue
- [`ImagePipelineDelegate`](https://kean-docs.github.io/nuke/documentation/nuke/imagepipelinedelegate) currently inherits from [`ImageTaskDelegate`](https://kean-docs.github.io/nuke/documentation/nuke/imagetaskdelegate) to allow to monitor all tasks. It can't be `@MainActor`.

There is another problem with this approach. Unfortunately, it breaks the invariant that `imageTaskCreated` is called immediately on the caller's thread and that can lead to logic issues. The way to avoid it is by making [`image(for:delegate:)`](https://kean-docs.github.io/nuke/documentation/nuke/imagepipeline/image(for:delegate:)) itself `@MainActor`, but that's not possible because you should be able to use it from any thread.

image

### Option 2

Use `nonisolated` when implementing `ImageTaskDelegate` methods. Unfortunately, it requires more code changes on the client side than just this:

Screen Shot 2022-06-27 at 8 20 42 PM

You can fix it by manually dispatching to the main actor, but it's bad.

```swift
nonisolated func imageTaskCreated(_ task: ImageTask) {
    Task { @MainActor in
        self.imageTask = task
    }
}
```

Unfortunately, it breaks the invariant that `imageTaskCreated` is called immediately on the caller's thread that can lead to logic issues. So this option is not feasible.

### Option 3

Wait until Swift introduced some sort of `@unchecked @MainActor` attribute to ignore these warnings – we know that the code is guaranteed to be called on the main queue.

### Option 4

Completely redesign this API. How?

### DIFF
--- a/Sources/Nuke/ImageTask.swift
+++ b/Sources/Nuke/ImageTask.swift
@@ -109,6 +109,7 @@ public final class ImageTask: Hashable, CustomStringConvertible, @unchecked Send
     }
 }
 
+// TODO: rework this
 /// A protocol that defines methods that image pipeline instances call on their
 /// delegates to handle task-level events.
 public protocol ImageTaskDelegate: AnyObject {


jefflewis:
Would making it an async sequence that returned "image task events", where that event could be one of progress, image preview or image be a valid alternative?

I think then the delegate could be done away with entirely.

kean:
@jefflewis, I'm considering this option. There can be two APIs: one that returns an image on completion and one that returns a sequence of events. But I'm not sure it solves the "return created task immediately" problem. Switching over events doesn't seem particularly elegant, but that can work unless there are other options.

bryansum:
+1 @jefflewis, that's what I first thought reading this. Enum for image image task events seems reasonable.

lvalenta:
First of all, you can always set the delegate to unchecked sendable without `MainActor`, no?

Also, do you have to return the `delegate`? Just before returning it, you can always map it to other (result) type that would have just `let` properties e.g.

jefflewis:
> One thing to note: the async sequence of events option seems to break down as soon as one of the delegate method needs to return a value (which is quite a common pattern for a delegate), that is it only works if the delegate is purely an events notification kind of thing.

How so? The sequence can be of any sendable type. So if the type that is being passed back is an enum with associated values that contains a Data, that could be the partial data or the fully downloaded image data. That could be an image already loaded from data.

tclementdev:
@jefflewis I meant you can't replicate this:

    protocol ImageTaskDelegate {
        func foo() -> SomethingHere
    }

kean:
If I were to go with the "async sequence" approach, I'm not sure how that sort of API will even look like. There still needs to be a base async/await API that simply returns a response.

```swift
public func image(for url: URL) async throws -> ImageResponse

/// ???
public func loadImage(with url: URL) -> AsyncThrowingStream<ImageTask.Event>
```

kean:
I'm going to proceed with an existing delegate-based approach. It works without warnings as long as you don't make your delegate a MainActor. If a new practice emerges or a new Swift feature becomes available to address it, I'll deprecate the existing API and introduce an improved one. But for now, the delegate is the least worst option.

PR #584, opened 2022-08-08
Add isVideoRenderingEnabled property for image

mbarashkov:
Currently LazyImage always creates an AVPlayer, as there is no way to set the isVideoRenderingEnabled property to it.
This pull request adds a isVideoRenderingEnabled parameter to LazyImage and does not initialize AVPlayer instances if it is set to false.

### DIFF
--- a/Sources/NukeUI/Image.swift
+++ b/Sources/NukeUI/Image.swift
@@ -73,7 +73,9 @@ public struct Image: UIViewRepresentable {
         if let resizingMode = self.resizingMode {
             imageView.resizingMode = resizingMode
         }
-        imageView.videoPlayerView.onVideoFinished = onVideoFinished
+		if imageView.isVideoRenderingEnabled {
+			imageView.videoPlayerView.onVideoFinished = onVideoFinished
+		}
         onCreated?(imageView)
         return imageView
     }
--- a/Sources/NukeUI/Image.swift
+++ b/Sources/NukeUI/Image.swift
@@ -18,18 +18,21 @@ public struct Image: NSViewRepresentable {
     let onCreated: ((ImageView) -> Void)?
     var onVideoFinished: (() -> Void)?
     var restartVideo: Bool = false
+	var isVideoRenderingEnabled: Bool = true
 
-    public init(_ image: NSImage) {
-        self.init(ImageContainer(image: image))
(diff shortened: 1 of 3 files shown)


kean:
Hi, @mbarashkov.

You can currently use [onCreated](https://kean-docs.github.io/nukeui/documentation/nukeui/image/init(_:oncreated:)) closure to achieve that:

```swift
Image(ImageContainer(image: .init())) {
    $0.isVideoRenderingEnabled = false
}
```

I missed one thing in the latest release though, it needs to be marked as a `MainActor`:

```swift
public init(_ imageContainer: ImageContainer, onCreated: @MainActor (ImageView) -> Void = { _ in })
```

As for the convenience API to achieve that, I think it should be something like this in SwiftUI:

```swift
NukeUI.Image(image).disableVideoRendering()
```

I'm absolutely for adding these convenience APIs, but I just didn't have time to work on them.

kean:
`LazyImage` also has an `onCreated()` closure which is supposed to be this escape hatch that allows full access to the underlying `UIKit` view.

Yes, it's probably also worth adding a `disableVideoRendering()` or `videoRenderingEnabled()` or some API like that.

`AVPlayer` is supposed to be initialized only when the video is about to be played. But I see there is a `AVPlayerLayer` type that is initialized eagerly. I think it could be made lazy too.

mbarashkov:
It creates video player in its constructor. And also in the display() method, I don't have guarantees it will not be called before the onCreate callback

kean:
I [added](https://github.com/kean/Nuke/commit/34fbc2a3108d1e9373e54c490076391863d60917) the following APIs to `Image`:

```swift
public func resizingMode(_ mode: ImageResizingMode) -> Self
public func videoRenderingEnabled(_ isEnabled: Bool) -> Self
public func videoLoopingEnabled(_ isEnabled: Bool) -> Self
public func animatedImageRenderingEnabled(_ isEnabled: Bool) -> Self
```

I also made sure `VideoPlayerView` is now created lazily when you use `LazyImage` or `Image`. And the underlying `AVPlayerLayer` is also created lazily.

For `LazyImage`, you'l need to use an initializer that works with `LazyImageState` and return a custom `Image` instance.

I think that should cover it.  I would appreciate if you could try it from the main branch. I'll release these changes soon.

mbarashkov:
@kean 
Works just fine; thanks.

PR #585, opened 2022-08-10
Update example code for LazyImageView

jeffreykuiken:
The example code for LazyImageView in the documentation still used the deprecated `source` attribute. This is a small fix to use the `url` attribute instead.

In the process of creating this PR I also came across two minor issues with the `onCompletion` closure, so I fixed those as well.


kean:
Thanks!

PR #599, opened 2022-09-10
Add Xcode 14 support

kean:


Sources/Nuke/Caching/ImageCache.swift
```diff
@@ -76,7 +76,7 @@ public final class ImageCache: ImageCaching {
     }
 
     /// Returns a recommended cost limit which is computed based on the amount
-    /// of the phisical memory available on the device.
+    /// of the physicalbenefitial memory available on the device.
```
own2pwn, line 79: physical

kean: Search & replace 👍

PR #602, opened 2022-09-16
Add support for loading image into TVPosterView

lukaskukacka:
`TVPosterView` now supports loading image using Nuke out-of-the-box (`TVPosterView` now conforms to `Nuke_ImageDisplaying`).

`TVPosterView` is a crucial UI component on tvOS.
While it is easy to add this conformance in your project, imho it is nice to have it out of the box in the library.

---

In my specific case, having it directly in the library is actually the only way to make my whole project to compile.

Custom conformance in my project worked fine with Nuke v. 10, but since update to v. 11 (which moves these extensions to `NukeExtensions`), using custom extension will no longer compile.

I guess it is caused by how Swift access control works + specific setup of my project (private frameworks with multi-level hierarchy + Nuke hidden privately in one of private frameworks under custom wrapper interface).

---

Thanks a lot for considering my change.
And big thanks for your time and effort on this amazing library! 🙂

### DIFF
--- a/Sources/NukeExtensions/ImageViewExtensions.swift
+++ b/Sources/NukeExtensions/ImageViewExtensions.swift
@@ -70,6 +70,17 @@ extension NSImageView: Nuke_ImageDisplaying {
 }
 #endif
 
+#if os(tvOS)
+import TVUIKit
+
+extension TVPosterView: Nuke_ImageDisplaying {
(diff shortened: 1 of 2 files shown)


kean:
Thanks for the contribution, @lukaskukacka. I appreciate that you even added a unit test.

PR #614, opened 2022-10-20
Update deprecated withTaskCancellationHandler

swasta:
Xcode 14.1RC shows warnings when withTaskCancellationHandler used with `handler` and `operation` arguments. This is a fix to call the renamed version without any warnings.


kean:
Looks good, thanks!

PR #615, opened 2022-10-20
add animated WebP asset type

makleso6:
Also see
https://developers.google.com/speed/webp/docs/riff_container#example_file_layouts
https://developers.google.com/speed/webp/docs/riff_container#extended_file_format


Sources/Nuke/Decoding/AssetType.swift
```diff
@@ -28,6 +28,8 @@ public struct AssetType: ExpressibleByStringLiteral, Hashable, Sendable {
     /// iOS 14, watchOS 7, tvOS 14.
     public static let webp: AssetType = "public.webp"
 
+    public static let animatedWebp: AssetType = "public.animated.webp"
```
kean, line 31: I don't think `"public.animated.webp"` is a thing.

I suggest adding something similar to [`ImageProperties`](https://github.com/kean/Nuke/blob/22ab51a030517fe2e170c4569b98d47fac4087a5/Sources/Nuke/Decoding/ImageDecoders%2BDefault.swift#L177) and potentially making the existing type public. It's currently internal because one of Nuke's goal is to keep the public API small.

kean:
> Nuke CI / Discover Typos (pull_request) failure

This is already fixed in the main branch.

makleso6:
I want to add support for animated WebP
but ```Animator``` and ```FrameStore```  is internal.
should I create protocol for ```Animator``` and ```FrameStore```?
In current implementation Animator works only with gif format

Idea - create protocol ```Animator``` and ```FrameStore```
implement ```GIFAnimator``` inherit from ``Animator``` and ```GifFrameStore``` inherit from  ```FrameStore```

In NukeWebP 
implement ```WebPAnimator``` inherit from ``Animator``` and ```WebPFrameStore``` inherit from  ```FrameStore```

@kean what do you think?

kean:
If you can reuse some of the parts of Gifu to add animated WebP support on top of it, that would be fantastic. I'm not familiar with its design. Ideally, please make as few changes as possible in the files under the /Gifu folder.

makleso6:
it seems that ```GIFImageView``` not using, probably should delete it?

kean:
> it seems that `GIFImageView` not using, probably should delete it?

I wouldn't because I'd like to keep it as close as possible to the source.

makleso6:
why is the source code of https://github.com/kaishin/Gifu located directly in the repository

kean:
It's embedded in order to avoid re-exporting its public APIs. It's meant to be a "private" dependency which are no yet supported.

kean:
I'm going to close it for inactivity. Please feel free to re-open if you to give it another go.

PR #626, opened 2022-11-17
Add new tests

zzmasoud:
Hey @kean 👋 I've added some new tests that will increase the code coverage from 91.3% to 92.3% .

### DIFF
--- a/Tests/NukeTests/ImageCacheTests.swift
+++ b/Tests/NukeTests/ImageCacheTests.swift
@@ -73,6 +73,36 @@ class ImageCacheTests: XCTestCase {
         // Then
         XCTAssertEqual(cache.countLimit, 1)
     }
+    
+    func testThatTTLChanges() {
+        //when
+        cache.ttl = 1
+        
+        // Then
+        XCTAssertEqual(cache.ttl, 1)
+    }
+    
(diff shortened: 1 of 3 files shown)


zzmasoud:
The previous failing check was because of `CoreImageFilter` and `GaussianBlur` on a WatchOS target which is unavailable. I've added `#if os(iOS) || os(tvOS) || os(macOS)` to prevent running these tests on that target.

zzmasoud:
So, the error is:

> testDefaultImageCost() encountered an error (Test runner never began executing tests after launching. If you believe this error represents a bug, please attach the result bundle at /Users/runner/Library/Developer/Xcode/DerivedData/Nuke-bycrsvaaztorfmesymoxfuvcgnfe/Logs/Test/Test-Nuke-2022.11.20_14-31-06-+0000.xcresult)
[1571](https://github.com/kean/Nuke/actions/runs/3495005634/jobs/5876739068#step:3:1572)

I've just checked and tested it again, and even with `randomized order ON`, nothing fails. I didn't modify that test at all.  Do you have any idea what's going on?

kean:
Thanks for the contribution, @zzmasoud. I apologize for the delay – I was on extended hiatus.

PR #637, opened 2022-12-20
`didComplete` edge cases.

0xceed:
Call `didComplete` even when images are cached or when tried to prefetch empty list.
@kean sorry, seems to be fixed now + handled edge case when tried to load empty list.

### DIFF
--- a/Sources/Nuke/Prefetching/ImagePrefetcher.swift
+++ b/Sources/Nuke/Prefetching/ImagePrefetcher.swift
@@ -107,24 +107,26 @@ public final class ImagePrefetcher: @unchecked Sendable {
     /// See also ``startPrefetching(with:)-1jef2`` that works with `URL`.
     public func startPrefetching(with requests: [ImageRequest]) {
         pipeline.queue.async {
-            for request in requests {
-                var request = request
-                if self._priority != request.priority {
-                    request.priority = self._priority
-                }
-                self._startPrefetching(with: request)
-            }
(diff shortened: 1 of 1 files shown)


Sources/Nuke/Prefetching/ImagePrefetcher.swift
```diff
@@ -107,32 +107,36 @@ public final class ImagePrefetcher: @unchecked Sendable {
     /// See also ``startPrefetching(with:)-1jef2`` that works with `URL`.
     public func startPrefetching(with requests: [ImageRequest]) {
         pipeline.queue.async {
-            for request in requests {
-                var request = request
-                if self._priority != request.priority {
```
0xceed, line 112: Removed this check: anyway we need all requests to has prefetcher's priority.

Sources/Nuke/Prefetching/ImagePrefetcher.swift
```diff
@@ -107,32 +107,36 @@ public final class ImagePrefetcher: @unchecked Sendable {
...
-                self._startPrefetching(with: request)
-            }
+            let starts = requests.map({
+                var copy = $0
+                copy.priority = self._priority
+                
+                return copy
+            }).map(self._startPrefetching(with:))
+            
+            guard starts.allSatisfy({ $0 == false }) else { return }
```
0xceed, line 117: `allSatisfy` returns `true` when called on empty list.

PR #667, opened 2023-02-08
Add the ability to render static gifs

Havhingstor:
The ImageView now tests if a gif is an animated gif, so that the animator only tries to animate it, if necessary. A static gif is rendered like any other image.

Previously, it was tried to animate every gif, whether it truly was animated or not. That resulted in no image being rendered at all.

### DIFF
--- a/Sources/NukeUI/Gifu/Classes/Animator.swift
+++ b/Sources/NukeUI/Gifu/Classes/Animator.swift
@@ -44,6 +44,15 @@ class Animator {
     return frameStore?.frameCount ?? 0
   }
 
+  /// Tests if the `data` is truly an **animated** gif
+  public static func isAnimatedGif(data: Data) -> Bool {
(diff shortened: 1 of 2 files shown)


kean:
Thanks, @Havhingstor. There were a couple of related issues reports, but nobody got to the bottom of this until now.

By the way, I'm considering removing the GIF support from the main framework entirely and directing users to integrate [Gifu](https://github.com/kaishin/Gifu) directly. The integration should only require a single line.

I think it's great to handle this scenario on the `ImageView` level for now. But I'd suggest also opening a similar PR in [Gifu](https://github.com/kaishin/Gifu) if you have time. I think it should be able to handle this scenario automatically.

kean:
I'm wondering how expensive `CGImageSourceCreateWithData` is going to be? Presumably not too expensive because it's not going to eagerly bitmap the image unless instructed to do so.

PR #671, opened 2023-02-14
async image loading task (sometimes) won't proceed when the task has been cancelled immediately

muukii:
## What happened

In my application, I found an issue that won't release resources captured by a task including image loading processing.
From a deep-dive investigation, I noticed something seems to be happening race-condition-ish inside ImagePipeline.

I described what's happening inside in the attached screenshot.

The below code is one of the cases that causes this case potentially.
(running in multi-thread a lot of times, reproducing is a bit hard)

```swift
let loadingTask = Task {
  let imageResponse = try await ImagePipeline.shared.image(for: request, delegate: nil)
  ...
  // there are some captured resources
  ...
}

...

loadingTask.cancel()
```

CleanShot 2023-02-15 at 01 33 03@2x

CleanShot 2023-02-15 at 01 34 14@2x

## Solution

I could fix this issue by adding a few lines of code that trigger cancel for the continuation even if the task is not found by the dictionary.
However, I'm not confident with this update. I guess should be added something more operations like handling delegates we could see in the original place. 
Or we should think about adding the task with mutex.

What do you think?
Thanks. I appreciate your great work.

### DIFF
--- a/Sources/Nuke/Pipeline/ImagePipeline.swift
+++ b/Sources/Nuke/Pipeline/ImagePipeline.swift
@@ -461,7 +461,10 @@ public final class ImagePipeline: @unchecked Sendable {
     }
 
     private func cancel(_ task: ImageTask) {
-        guard let subscription = tasks.removeValue(forKey: task) else { return }
+        guard let subscription = tasks.removeValue(forKey: task) else {
+            task.onCancel?()
+            return
+        }
         dispatchCallback(to: task.callbackQueue) {
             if !task.isDataTask {
                 self.delegate.imageTaskDidCancel(task)


Sources/Nuke/Pipeline/ImagePipeline.swift
```diff
@@ -461,7 +461,10 @@ public final class ImagePipeline: @unchecked Sendable {
     }
 
     private func cancel(_ task: ImageTask) {
-        guard let subscription = tasks.removeValue(forKey: task) else { return }
+        guard let subscription = tasks.removeValue(forKey: task) else {
+            task.onCancel?()
```
muukii, line 465: still just prototyping code

kean:
Hey, @muukii. 

For most scenarios, this code looks good. The `startImageTask` is added to the internal serial dispatch queue before `loadImage` method returns, so there is no chance to cancel the task before that. I do see one potential issue though. I've recently added these new delegate methods – `imageTaskCreated` and `imageTaskDidStart` – that get called before the underlying task is created and registered. If you happen to call `cancel` in one of these methods, that will most certainly lead to an issue.

muukii:
@kean 
Thank you for replying.

> For most scenarios, this code looks good. 

meaning, the current code is good? or my draft code is looking good?
I hope this case will be fixed. I think this is probably frequently happening in case of image-loading is running or stopping according to scrolling in list view.

kean:
I meant the existing code in the framework, and the way it interacts with the dispatch queues. It should fork fine except for the scenario I outlined.

The suggested change also looks good and I tested it too. I added the following unit test in `ImagePipelineAsyncAwaitTests.swift`:

```swift
    func testCancelFromTaskCreated() async throws {
        // GIVEN
        taskDelegate.onTaskCreated = { $0.cancel() }

        let task = Task {
            try await pipeline.image(for: Test.url, delegate: taskDelegate)
        }

        var caughtError: Error?
        do {
            _ = try await task.value
        } catch {
            caughtError = error
        }
        XCTAssertTrue(caughtError is CancellationError)
    }
```

Before I apply the changes, it hangs. With the changes, it succeeds.

> I think this is probably frequently happening in case of image-loading is running or stopping according to scrolling in list view.

My understanding is that it should never happen unless the task gets immediately cancelled from `imageTaskCreated`.

Let's make sure we identify the issue first. Can you please provide a sample where you could reproduce it? Does the task get cancelled in `imageTaskCreated`? Or are there more scenarios we need to address? It's a bit hard to reason about this code with so many public APIs. I want to make sure everything is covered and the solution is sound.

muukii:
I created a repo that reproduces this case.
https://github.com/muukii/Nuke-reproduce-hang

To debug resource allocating and deallocating, use this

```swift
final class Resource {
  
  init() {
    print("Init", self)
  }
  
  deinit {
    print("Deinit", self)
  }
}
```

```swift
let r = Resource()
        
let task = Task {
  
  let response = try await ImagePipeline.shared.image(for: URL(string: "https://images.unsplash.com/photo-1675097659737-2575571c507c?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=2940&q=80")!,delegate:nil)
  
  print(response)

  withExtendedLifetime(r) {}
}

task.cancel()
```

hit the button, then `deinit` won't be printed

kean:
Hey, @muukii. I really appreciate you creating a sample. Now the issue is clear, and it's a different one from the one from my test case.

I tested it in Xcode 14.2 and saw the following order of calls (which is not exactly what I would've expected).

image

I'm trying to think how to best solve it. I think what you suggested is most likely OK, but because there is multi-threading involved, I'm not entirely sure yet.

kean:
I think I have a solution. By going through the documentation and a couple of forums [posts](https://forums.swift.org/t/how-to-use-withtaskcancellationhandler-properly/54341/15), it's clear that `operation:` and `onCancel:` handlers can be called in parallel and the order of calls is undefined. It means that the pipeline has to take care of synchronization. Fortunately, it seems to be fairly easy to do:

Screenshot 2023-02-19 at 8 46 41 PM

- If `onCancel:` is called before `operation:`, `task.state` will be guaranteed to be `.cancelled` by the time we reach `withUnsafeThrowingContinuation` closure.
- If `onCancel:` is called in parallel with `operation:`, e.g. at the same time as line 153 is executing, the actual work to cancel will be dispatched to the pipeline's serial queue, and it'll safely cancel.
- If `onCancel:` is called after the task is started, it's just a regular cancellation scenario – no issue here.

I'm going to update `data(for:)` methods too – they have the same issue.

It bit me again that the pipeline is not guaranteed to call the completion closure if you cancel the task. But it's been true since the inception of the framework. It might be worth revising it in the next major release. It would've made it easier designing around these things.

kean:
I release a fix in [version 11.6.4](https://github.com/kean/Nuke/releases/tag/11.6.4).

PR #696, opened 2023-05-03
Fix #695: byCroppingToSquare() always return square image

### LINKED ISSUE #695: ios 11 crash
I find that when the source image's width bigger than height just 1 pixel, then after called cropping, the result is same to source image. so, image's width is not equal to the height. we should use min(width, height) as radius. Can you fix it?
image

zzmasoud:
Fixed #695 by rounding down the offset calculated in:
https://github.com/kean/Nuke/blob/f854af84cd8c35635bce2c3e3a3eb366fc6db446/Sources/Nuke/Internal/Graphics.swift#L72

The issue is all about fractional digits error (0.5), and we don't have 0.5 pixel, so it's whether 0 or 1. I decided to go with rounding down.
Now, it's guaranteed that the output image is always square.


kean:
I appreciate it, thanks!

PR #701, opened 2023-06-02
Improve random number generation

woxtu:
Hi,

This PR replaces legacy random API with new API to improve random number generation.


kean:
That's much nicer, thanks!

Issue #693, opened 2023-04-14
DataPublisher created from async function should start lazily


alexvasilkov:
I need to show a photo grid where each photo has a tiny preview stored in local DB, and I want to display these previews shortly until a bigger thumbnail is loaded from remote URL.

To load local previews I'm creating a custom `ImageRequest` like this:

```swift
    let request = ImageRequest(
      id: "preview_\(photoId)",
      data: { await loadPreviewFromDb(photoId) },
      options: [.disableDiskCache]
    )
```

Now when the grid is shown only the first 6 previews are loaded and the rest are just ignored, even though I can confirm that the actual data is loaded from the DB correctly. In the logs I see `Data loader returned empty data.` errors.

Adding `.skipDataLoadingQueue` option kinda helps but I still see random blank previews from time to time.

It turns out the async `data` function is called eagerly when it's converted into `Publisher` (see [DataPublisher.swift/publisher(...)](https://github.com/kean/Nuke/blob/6de55b6c3e67e95c02dc8a1521c1dd02f4c2a3ea/Sources/Nuke/Internal/DataPublisher.swift#L36) code), and `PassthroughSubject` will ignore data that was sent while it had no subscribers (queued?).

I managed to fix it by using lazy `Publisher` created like this:

```swift
    let request = ImageRequest(
      id: "preview_\(photoId)",
      dataPublisher: Future { await loadPreviewFromDb(photoId) },
      options: [.disableDiskCache]
    )
```

```swift
private extension Future where Failure == Error {
  convenience init(action: @escaping () async throws -> Output) {
    self.init { promise in
      Task {
        do {
          let result = try await action()
          promise(.success(result))
        } catch {
          promise(.failure(error))
        }
      }
    }
  }
}
```

This seems to work just fine, but I think it worth fixing it in the library as well.

kean:
Nice catch; thanks for the report. It was never intended to work this way. The async work should be performed lazily. The implementation-wise, it should probably not perform the conversion to a publisher, but it's also OK if it does. There are ways to make it lazy. I'll address it when I get some time, but in the meantime, PRs are always welcome.

alexvasilkov:
Sorry for delayed reply, I wanted to test it in production to make sure it works, and it doesn't work that good after all.

About 0.5% of users who received this update got crashes like this:

```
0  libswiftCore.dylib             0x39c40c _swift_release_dealloc + 32
1  Combine                        0x1ded4 Future.__deallocating_deinit + 204
[...]

PR #708, opened 2023-06-22
#693: DataPublisher created from async function should start lazily

khlopko:
Fix for the issue: https://github.com/kean/Nuke/issues/693.

`Future` is generally better approach than `PassthroughSubject` when you need to execute something once (e.g. network requests). What was failing in the original solution by the issue author is that `Future` starts execution right after creation as well, and needs to be wrapped in `Deferred` in order to start only after subscribing to it.


kean:
Works as advertised, thanks, @khlopko. I appreciate you taking time to address it.

PR #711, opened 2023-06-26
Fix NukeUI errors for watchOS

### LINKED ISSUE #710: ImageProcessors.GaussianBlur unavailable on watchOS makes building NukeUI for watchOS fail
Hi, the usage of [ImageProcessors.GaussianBlur in the demo of LazyImage](https://github.com/kean/Nuke/blob/main/Sources/NukeUI/LazyImage.swift#L249) causes building NukeUI (& apps that want to use NukeUI) for watchOS fail. I would PR to fix, but unsure how best you'd want to fix this. I would just add a conditional checking if it's on watchOS and if it is don't include that processors modifier, but then the demo wouldn't blur when it's meant to.

FieryFlames:
Resolves #710
Material.ultraThick is available on watchOS 10 and up, but it seemed like too much work to conditionally apply it

### DIFF
--- a/Sources/NukeUI/LazyImage.swift
+++ b/Sources/NukeUI/LazyImage.swift
@@ -246,7 +246,9 @@ private struct LazyImageDemoView: View {
                     image.resizable().aspectRatio(contentMode: .fit)
                 }
             }
+#if !os(watchOS)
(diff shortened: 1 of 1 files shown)


kean:
Thanks!

PR #713, opened 2023-07-05
Remove useless extension

openmetrue:
Removed unnecessary extension, because in iOS 8.0/macOS 10.10/tvOS 9.0/watchOS 2.0 it is already in the Foundation.
https://developer.apple.com/documentation/foundation/nslocking/4059821-withlock

All tests on all platforms were successful.

### DIFF
--- a/Sources/Nuke/Caching/Cache.swift
+++ b/Sources/Nuke/Caching/Cache.swift
@@ -20,8 +20,8 @@ final class Cache<Key: Hashable, Value>: @unchecked Sendable {
     }
 
     var conf: Configuration {
-        get { lock.sync { _conf } }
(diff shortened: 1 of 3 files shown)


kean:
Wow, I just now learned about it. Thanks a lot!

openmetrue:
> Wow, I just now learned about it. Thanks a lot!

Just be careful, this support appeared in xcode 14

kean:
Yes, no issue at all.

> Please note, as of April 2023, all iOS and iPadOS apps submitted to the App Store must be built with a minimum of Xcode 14.1 and the iOS 16.1 SDK

I'll update the minimum supported Xcode version to 14.1 and update the CI, unless you beat me to it.

PR #715, opened 2023-07-14
Minor docs correction in FetchImage

tdkn:
This pull request addresses a property name issue within FetchImage. Previously,
the code referenced a `view` property, which does not exist.
This has now been corrected to `image`, aligning with the existing properties within FetchImage.

Thank you again for your incredible work and commitment to this project.

CleanShot 2023-07-14 at 10 08 53


kean:
Nice! I always appreciate the updates to the documentation. There is a lot of it, and it's always hard to keep it up to date if you are not actively using it yourself.

I keep a test file with some, but not all, of the executable code from the documentation to make sure it still compiles: https://github.com/kean/Nuke/blob/main/Tests/NukeTests/ImagePipelineTests/DocumentationTests.swift, but apparently this sample isn't part of it.

PR #735, opened 2023-11-14
Add another file type signature for .m4v files

leonid-shevtsov:
Hi, the `AssetType` file type detection code is missing one of the M4V signatures I've seen in the wild.

I've found a [catalog of signatures](https://www.garykessler.net/library/file_sigs.html) that confirms that.

HTH


kean:
LGTM, thanks!

I knew Railsware sounded familiar. I lived in Szlak 50/A4 for a month and saw their offices every day 😆

PR #736, opened 2023-11-16
Added the onStart callback to SwiftUI.LazyImage

urbaneewe:


Sources/NukeUI/FetchImage.swift
```diff
@@ -184,6 +188,10 @@ public final class FetchImage: ObservableObject, Identifiable {
                 handle(result: .failure(error))
             }
         }
+
+        if let task = imageTask {
```
kean, line 192: The task is always going to be empty at this point, to it's not going to get called. I'm not entire sure why I added this `load()` variant in the first place, to be honest, but I think it would be acceptable to not call `onStart` here.

urbaneewe: Yup, I agree, I wasn't sure if you had something planned so I added it just in case. I'll remove it

kean:
Hey, thanks for fixing the typos as well, @urbaneewe; merging.

PR #749, opened 2024-01-18
Pass alphaInfo from CGImage to CGContext during decompression

dmytro-kasianchuk-doordash:
Partially transparent images sometimes are improperly rendered with decompression on. It's somewhat difficult to reproduce as it requires using partially transparent images and rendering multiple images at the same time.

It looks like Nuke is defaulting to CGAlphaInfo `.premultipliedLast` which causes an issue. CGImage has Alpha information, passing it fixes the issue.

### DIFF
--- a/Sources/Nuke/Internal/Graphics.swift
+++ b/Sources/Nuke/Internal/Graphics.swift
@@ -134,7 +134,7 @@ extension PlatformImage {
         guard let cgImage = cgImage else {
             return nil
         }
-        guard let ctx = CGContext.make(cgImage, size: canvasSize) else {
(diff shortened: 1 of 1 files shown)


dmytro-kasianchuk-doordash:
Hey @kean, it looks like my PR is effectively undoing https://github.com/kean/Nuke/commit/ff967f4bb6083f9dc7cff616dcc3b6202ecc8e1e. 

I don't have context on the original fix and I've tried to repro the issue in an isolated project to no success. I'm closing this for now. When I understand the exact repro conditions I'll revisit this

PR #743, opened 2023-12-25
Add support for visionOS

bobek-balinek:
- Introduced a versioned Package.swift as per [specification](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md\#version-specific-manifest-selection)
- Added `.visionOS` as a supported platform
- Updated handling of the background mode of AVPlayer for visionOS
- Disabled `ImageDecoders.Video` as `AVAssetImageGenerator.copyCGImage` is not available for visionOS
- Resolves #706

PR #752, opened 2024-02-08
Enable visionOS support for all APIs

zachwaugh:
Package support for visionOS was enabled in https://github.com/kean/Nuke/pull/743, but almost all the APIs were not specifically enabled for visionOS, so it wasn't really usable when updating my iOS app to run natively on visionOS. This enables visionOS everywhere iOS is enabled with a single exception (`UIScreen` isn't available on visionOS).

### DIFF
--- a/Sources/Nuke/Caching/Cache.swift
+++ b/Sources/Nuke/Caching/Cache.swift
@@ -4,7 +4,7 @@
 
 import Foundation
 
-#if os(iOS) || os(tvOS)
+#if os(iOS) || os(tvOS) || os(visionOS)
 import UIKit.UIApplication
 #endif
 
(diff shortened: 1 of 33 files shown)


kean:
Thanks, @zachwaugh. I appreciate you taking the time to put a PR together.

zachwaugh:
@kean no problem, thanks for the library! The one thing I didn't get a chance to add here though was enabling visionOS for all the Xcode project targets as well as in CI. I can't take a look at that when I get some time unless you or someone else gets to it sooner.

kean:
>  enabling visionOS for all the Xcode project targets as well as in CI

Yeah, they're still a few more steps, including updating the scripts that create xcframework binaries. I wanted to tackle it, but it looks like you need an Xcode beta to do it, which I don't have on my machine.

Issue #750, opened 2024-01-18
storeCachedData: callback once stored in cache


AndrewSB:
hey, i was using the 2 APIs `ImagePipeline.shared.cache.storeCachedData(imageData, for: ImageRequest(url: remoteURL)` + `ImagePipeline.shared.loadImage(with: url)`.

i'm using them in pretty quick succession, when a user captures an image that is being uploaded. i'd like to await the completion of `storeCachedData` before i allow the call to `loadImage`, but afaict there's no callback/notification once the cached image has been saved. the 2 objects that make the calls are quite disconnected, and i'd prefer to not have them depend on oneanother.

how can i ensure that i don't make that `loadImage` call until `storeCachedData` has made the imageData available in the cache?

kean:
If you are using the built-in `DataCache`, it has a so-called staging area where it stores the data while it's being saved on disk asynchronously, so there is no need to wait until it finishes. It's a documented behavior, so you can rely on it continuing working this way.

AndrewSB:
i believe i am, but i'm noticing a slight delay. my flow is

1. cache the image
2. in parallel, fetch the image and show it in a view that's presented modally

the image shows up black (i.e., i think it fails to load).

if i dismiss the modal, then re-present it, i think it works. so likely, there's something with threading/waiting for the staging area to accept the value before i try to fetch it

is there some way i can syncronize that? by submitting the logic to present the modal to a certain thread, or waiting for a notification to fire?

kean:
Hey, can you please share the code that you are using? 

There is a public API: [`flush(for:)`](https://kean-docs.github.io/nuke/documentation/nuke/datacache/flush(for:)) that clears the staging area and puts the data on disk immediately. But there should be no need in using it, so I'm not sure it's going to help in this case. The root cause is likely somewhere else.

AndrewSB:
https://github.com/kean/Nuke/pull/755 is how i'm using the API. from running the test i just wrote, it looks like it doesn't result in a cache hit

AndrewSB:
solved by the PR

PR #755, opened 2024-02-21
ThreadSafetyTests: add failing test where we load and store an image

AndrewSB:
adds a test that shows https://github.com/kean/Nuke/issues/750

### DIFF
--- a/Tests/NukeThreadSafetyTests/ThreadSafetyTests.swift
+++ b/Tests/NukeThreadSafetyTests/ThreadSafetyTests.swift
@@ -164,6 +164,31 @@ class ThreadSafetyTests: XCTestCase {
         }
         queue.waitUntilAllOperationsAreFinished()
     }
+
+    func testDataCacheMultipleThreadAccess() {
+        let aURL = URL(string: "https://example.com/image-01-small.jpeg")!
+        let imageData = Data(repeating: 1, count: 256 * 1024)
+
+        let expectSuccessFromCache = self.expectation(description: "one successful load, from cache")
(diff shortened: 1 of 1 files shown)


kean:
Hey @AndrewSB, thanks for providing the sample code. I now realized what I was missing, and it was pretty obvious.

The default pipeline (`ImagePipeline.shared`) does not use `Nuke.DataCache` but instead relies on the `Foundation.URLCache`. Unfortunately, the following APIs only with the custom caches added by Nuke and not the caches that are part of the data loading system:

```swift
ImagePipeline.shared.cache.storeCachedData(imageData, for: ImageRequest(url: remoteURL)
```

In order to make this test pass, you need to do the following:

```swift
// 1. Enable custom disk cache
let pipeline = ImagePipeline(configuration: .withDataCache)

// 2. Use real image data, otherwise the cached data gets ignored
let imageData = try XCTUnwrap(Test.data(name: "fixture", extension: "jpeg"))
```

With these two changes, you get the following response:

```
    ▿ cacheType : Optional<CacheType>
      - some : Nuke.ImageResponse.CacheType.disk
```

Caching is one of the more compiled aspects in any image loading framework, so I would recommend going through the related [documentation](https://kean-docs.github.io/nuke/documentation/nuke/cache-layers).

kean:
Thanks for updating the test. I'm going to merge and update it a bit – it shouldn't use the default cache.

AndrewSB:
that makes sense! i wrote it that way because that's how i'm using it in my project, i assumed you'd want on which is created inline. appreciate the help @kean!

PR #762, opened 2024-03-06
Apply if let shorthand syntax

mlight3:
if let shorthand syntax was introduced since swift 5.7. ([SE-3045](https://github.com/apple/swift-evolution/blob/main/proposals/0345-if-let-shorthand.md))

I replace codes by using following regexs.

regex | replacement
--- | --- 
if let (.+)? = \1 \{ | if let $1 {
if let (.+)? = \1, | if let $1,
if let (.+)? = self.\1 \{ | if let $1 {
if let (.+)? = self.\1, | if let $1,


kean:
Nice work, thank you @mlight3. 

I noticed there is also at least one `var` that hasn't been updated, in case you'd want to tackle it as well:

```swift
guard var request = request else {
```

mlight3:
Hi ,@kean 

Thank you for the review.
I also replaced `guard statement` in this [commit](https://github.com/kean/Nuke/pull/762/commits/2b32e6771e5de69d8758f0d95d4065af4888d991)

PR #764, opened 2024-03-07
Add usePlaceholderAsFailure to show placeholder in case of image loading failure

mlight3:
I added `usePlaceholderAsFailure` because I think there are a lot of cases where people might want to keep the placeholder visible even if the image fails to load.


mlight3:
@kean The default value of `usePlaceholderAsFailure` is `true`, do you think it should be a `false`?

kean:
I changed the default behavior to `false` and renamed it a bit. Thank you for the contribution!

Issue #765, opened 2024-03-07
Xcode 15.3 concurrency warnings when using `Screen.scale`


jszumski:
Building with Xcode 15.3 shows 2 new concurrency warnings:

> Sources/Nuke/Processing/ImageProcessors+Resize.swift:88:60 
> Main actor-isolated static property 'scale' can not be referenced from a non-isolated context; this is an error in Swift 6

> Sources/Nuke/Internal/Graphics.swift:198:44
> Main actor-isolated static property 'scale' can not be referenced from a non-isolated context; this is an error in Swift 6

Screen is indeed marked `@MainActor`:

```
@MainActor
enum Screen {
#if os(iOS) || os(tvOS)
    /// Returns the current screen scale.
    static let scale: CGFloat = UIScreen.main.scale
#elseif os(watchOS)
    /// Returns the current screen scale.
    static let scale: CGFloat = WKInterfaceDevice.current().screenScale
#else
    /// Always returns 1.
    static let scale: CGFloat = 1
#endif
}
```

PR #766, opened 2024-03-07
Fix Xcode 15.3 concurrency warnings when using `Screen.scale`

jszumski:
Fixes https://github.com/kean/Nuke/issues/765.  The viral nature of concurrency unfortunately means that the blast radius is pretty high for just 2 warning fixes:

- Adds `@MainActor` to direct and downstream callers of `Screen.scale`
- Adds `@MainActor` to downstream callers in tests
- Moves `@MainActor` from XCTestCases to individual test functions instead
- Switches the image pipeline thread safety test from resizing to blurring.  This avoids marking the test as `@MainActor` which would destroy its ability to test threading.

### DIFF
--- a/Sources/Nuke/Internal/Graphics.swift
+++ b/Sources/Nuke/Internal/Graphics.swift
@@ -314,7 +314,6 @@ extension CGSize {
     }
 }
 
-@MainActor
 enum Screen {
 #if os(iOS) || os(tvOS)
     /// Returns the current screen scale.


kean:
Hey, @jszumski. Thanks for raising the issue and taking a stab at it.

> means that the blast radius is pretty high for just 2 warning

Right, I also would prefer to avoid limiting most of these APIs to the main-thread. What I would suggest trying instead is updating [`ImageTargetSize`](https://github.com/kean/Nuke/blob/058bf6c5bd9aec75c9be5fe16f287997b0683b9f/Sources/Nuke/Processing/ImageProcessors%2BResize.swift#L80) so that it no longer pre-computes the size in pixels.

jszumski:
> What I would suggest trying instead is updating `ImageTargetSize` so that it no longer pre-computes the size in pixels.

Any pointers to where an async call to the MainActor might fit best?  I don't know much about the internals of Nuke, but I assume the entire pipeline should be able to run in a background thread, so it would have to be done in an initial step somewhere and fed down into these other call sites?

kean:
>  Any pointers to where an async call to the MainActor might fit best?

I briefly looked into it, and I'm really not sure. None of these potential solutions seem ideal. You need the pixel size during resizing and resizing happens in the background. I'm also not sure how pre-computing it would look like. After considering these options, restricting certain, but not all, initializers to the main thread doesn't seem like the worst option.

kean:
If you remove the `@MainActor` attribute from `Nuke.Screen`, the warning goes away. It was added in 2022 to fix a warning in one of the previous versions of Xcode.

kean:
Yeah, it seems like a compiler bug. I gave up trying to find any official documentation on how it's supposed to work. The rules for Sendable will now probably require a separate book.

_Maybe_ it  works because it's `static` and the compiler knows to init it on the main actor.

kean:
> Updated with that fix. I'm honestly surprised it works.

I think it was just a bug with "minimum" concurrency checking, which I use, and now it's been fixed. I wouldn't expect a warning here with "minimum" checking.

Btw, if you enable "complete" checking, it becomes an error:

Screenshot 2024-03-26 at 7 53 12 PM

jszumski:
I wonder if `UITraitCollection.current.displayScale` is a better replacement that could sidestep the main actor requirement?

kean:
> I wonder if UITraitCollection.current.displayScale is a better replacement that could side step the main actor requirement?

Great suggestion! It looks like it is and it is not confined to the main thread:

```
UIKit stores the value of the currentTraitCollection property as a thread-local variable, so access is lightweight and free of side effects. 
```

I'm going to go with `UITraitCollection.current.displayScale` in the upcoming version.

PR #768, opened 2024-03-09
Fix image loading test

woxtu:
This PR fixes `ImageViewIntegrationTests.testLoadingWithNilURL()` because the test has failed.
Loading an image from an invalid URL should return a data loading error?


woxtu:
Hmm...

kean:
In this scenarios, the URL is `nil`, so it throws `.imageRequestMissing`.

woxtu:
I figured out the problem. "http://example.com/invalid url" is a valid URL in Xcode 15, so the test fails.

woxtu:
I figured out the problem. In Xcode 15, "http://example.com/invalid url" is a valid URL so the test fails.

PR #769, opened 2024-03-10
Fix thumbnail key for image request

woxtu:
This PR fixes the raw value of `ImageRequest.UserInfoKey.thumbnailKey` to match the format with other keys.


kean:
It's only used in memory, so I think it's a safe change 👍

PR #773, opened 2024-03-19
Add support for configurable placeholders on `LazyImage`

zenangst:
Hey @kean, we used `Nuke` for all image loading in our app, so first of all, thanks for making such an amazing framework. One thing that we noticed is that we can sometimes end up with the colored default view when loading images. So with this PR, the consumer of the framework can now decide what view should be used as a placeholder instead of the default color that the framework provides. What do you think?

By using default arguments for the newly added placeholder closure, we can keep the public API intact and offer more functionality without causing any issues for existing users.

Hope you like it. 

## Summary

Add additional placeholder closure to the public init methods so that the consumer of the framework can decide what view to use as a placeholder instead of defaulting to `Color(.secondarySystemBackground)`. The public API remains unchanged as the default argument is still `Color(.secondarySystemBackground)`

### DIFF
--- a/Sources/NukeUI/LazyImage.swift
+++ b/Sources/NukeUI/LazyImage.swift
@@ -17,7 +17,7 @@ public typealias ImageRequest = Nuke.ImageRequest
 /// task coalescing, smart background decompression, request priorities, and more.
 @MainActor
 @available(iOS 14.0, tvOS 14.0, watchOS 7.0, macOS 10.16, *)
-public struct LazyImage<Content: View>: View {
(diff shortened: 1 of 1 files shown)


zenangst:
@kean It took me a while to figure out where the color actually came from when we got a report on this issue, hence us starting to use the other initializer. I just thought it would be a good idea for the user to be able to specify the default placeholder so that they could pick the most suitable initializer.

Our app uses a lot of colors, so the designers had a field day when the pointed it out to us 😆

PR #785, opened 2024-05-28
Update task description

ejensen:
Now that [Pulse uses task descriptions](https://github.com/kean/Pulse/pull/254), the Pulse console shows all Nuke network requests as the hard-coded `Nuke Load Data` value. 
~This PR appends the request description to the task description to differentiate the entries~ 
This PR removes the task description assignment.


kean:
I have no memory of this code 🤔 
I would probably remove the description entirely.

PR #792, opened 2024-06-17
Fix “unrecognized selector sent to instance” crash in transition on iOS < 17

ejensen:
Prevents an “unrecognized selector sent to instance” crash caused by UIKit incorrectly marking [`UIImageIVew.preferredImageDynamicRange`](https://developer.apple.com/documentation/uikit/uiimageview/4173133-preferredimagedynamicrange) available since iOS 2.0. Its actual availability is iOS 17.0+

The reference to `preferredImageDynamicRange` was added in https://github.com/kean/Nuke/pull/787 and released in Nuke v12.7.1

Incorrect documentation

### DIFF
--- a/Sources/NukeExtensions/ImageViewExtensions.swift
+++ b/Sources/NukeExtensions/ImageViewExtensions.swift
@@ -413,9 +413,9 @@ extension ImageViewController {
         transitionView.frame = imageView.frame
         transitionView.tintColor = imageView.tintColor
         transitionView.tintAdjustmentMode = imageView.tintAdjustmentMode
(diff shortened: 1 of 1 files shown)


kean:
Yeah, it looks like a mistake in the label. The type that it returns, [`UIImage.DynamicRange`](https://developer.apple.com/documentation/uikit/uiimage/dynamicrange) is labeled as iOS 17+. Let's get this merged, or is there some issue?

kean:
** TEST BUILD FAILED ** on Xcode 14.3.1. I'm going to take care of it in the main branch. I think it time to drop its support because Xcode 15 has been a requirement for shipping to App Store since April.

Issue #791, opened 2024-06-13
Image scaling bug


prabhuamol:
We are sometimes noticing when images are scaled using .thumbnailKey they appear blurry. I wonder if its cause in` func makeThumbnail(data: Data, options: ImageRequest.ThumbnailOptions) -> PlatformImage?` scale is not taken into account when returning back the UIImage. Which is causing the dimensions of the image to be reported incorrectly.

We do some post processing via `prepareThumbnail` if the returned thumbnail size does not match the view size exactly and the scaled down image returned using https://developer.apple.com/documentation/uikit/uiimage/3750845-preparethumbnailofsize is too blurry.

Heres a sample request and response to show the issue

▿ ImageRequest(resource: "some url (hidden for privacy)", priority: normal, processors: [], options: Options(rawValue: 0), userInfo: [Nuke.ImageRequest.UserInfoKey(rawValue: "github.com/kean/nuke/thumbnail"): Nuke.ImageRequest.ThumbnailOptions(targetSize: Nuke.ImageRequest.ThumbnailOptions.TargetSize.fixed(474.0), createThumbnailFromImageIfAbsent: true, createThumbnailFromImageAlways: true, createThumbnailWithTransform: true, shouldCacheImmediately: true), Nuke.ImageRequest.UserInfoKey(rawValue: "imageSize"): (110.0, 158.0)])
  ▿ ref : <Container: 0x600001819c40>
  
 
(lldb) po response.image
<UIImage:0x6000030307e0 anonymous {338, 474} renderingMode=automatic(original)>

(lldb) po response.image.size
▿ (338.0, 474.0)
  - width : 338.0
  - height : 474.0

As you can see image size is reported in pixel size and not points. The fix might be that when we return PlatformImage in Nuke.makeThumbnail use scale initializer for UIImage. https://github.com/kean/Nuke/blob/084076db4f9b9ee4db51177ef5a4879689c1aa09/Sources/Nuke/Internal/Graphics.swift#L369

`return PlatformImage(cgImage: image, scale: scale, orientaion: orientation)` instead of return `PlatformImage(cgImage: image)` I had to do that in an app i was working on as well

prabhuamol:
You can also verify this by just chaning any of the `ImageProcessorsResizeTests` to use points and all of them fail

[...]

PR #793, opened 2024-06-21
Fix image scale issue in thumbnail creation

prabhuamol:
Potential fix for https://github.com/kean/Nuke/issues/791 Also makes sure resized images have the same orientation as the original image

Also renamed the image which was right oriented correctly


Sources/Nuke/Decoding/ImageDecoders+Default.swift
```diff
@@ -52,7 +52,9 @@ extension ImageDecoders {
 
             func makeImage() -> PlatformImage? {
                 if let thumbnail {
-                    return makeThumbnail(data: data, options: thumbnail)
+                    return makeThumbnail(data: data,
```
prabhuamol, line 55: use scale in creating a thumbnail similar to how `ImageDecoders.Default._decode(data, scale: scale)` is using it below

Sources/Nuke/Internal/Graphics.swift
```diff
@@ -366,7 +383,22 @@ func makeThumbnail(data: Data, options: ImageRequest.ThumbnailOptions) -> Platfo
     guard let image = CGImageSourceCreateThumbnailAtIndex(source, 0, options as CFDictionary) else {
         return nil
     }
+    
+#if canImport(UIKit)
+    let orientation: UIImage.Orientation
```
prabhuamol, line 388: this is needed to make sure the returned image has the same orientation as the original. Verified via tests below with .right and .up orientation images

kean:
Hey, thanks @prabhuamol. I think it looks really good now.

kean:
I will update the tests in `main`.

PR #802, opened 2024-07-28
Nuke 13 – Swift 6

kean:
## Requirements

- Xcode 16

### ImagePipelineActor

Nuke has been [designed](https://kean.blog/post/concurrency) around the global actor pattern for quite a while now, so it was an easy change to introduce a global `ImagePipelineActor` as a replacement. `ImagePipeline`, `ImagePrefetcher`, `ImageTask`, and all `AsyncTask` implementations are now syncrhonized on the new actor instead of a a serial GCD queue.


jcbriones (approved):
Lots of detailed updates and thank you Alexander for adding this full swift6 support.

Sources/Nuke/Internal/Mutex.swift
```diff
@@ -4,11 +4,11 @@
 
 import Foundation
 
-final class Atomic<T>: @unchecked Sendable {
```
larryonoff, line 7: I assume that it would be good to take [Mutex](https://github.com/pointfreeco/swift-sharing/blob/main/Sources/Sharing/Internal/MutexBackport.swift) backport implementation from `swift-sharing` framework, which is ported from `Swift.Synchronization`

kean: Hmm, interesting. Thanks for the link. I'll look into it.

laraknutson:
Any updates for this fix?

PR #806, opened 2024-08-08
Fix a leak related to self not being weak

prabhuamol:
Fix for issue submitted here https://github.com/kean/Nuke/issues/807


Sources/Nuke/Tasks/TaskFetchOriginalData.swift
```diff
@@ -54,7 +54,8 @@ final class TaskFetchOriginalData: AsyncPipelineTask<(Data, URLResponse?)> {
                 guard let self else {
                     return finish()
                 }
-                self.pipeline.queue.async {
+                self.pipeline.queue.async { [weak self] in
```
kean, line 57: There isn't much point for briefly retaining it during the thread hop. `weak` references incur a cost as they have to spin lock to count references. 

I'm rewriting this code in [Nuke 13](https://github.com/kean/Nuke/pull/802). It will be all confined to the same global actor.

PR #815, opened 2024-09-22
execute Task on ImagePipeline DispatchQueue using executorPreference

georgesjamous:
The ImagePipeline is currently creating a Task and then immediately switching to its internal queue. This results in unnecessary context switching from the Task’s originating queue to the pipeline’s internal queue. 

So, to optimize this and reduce this overhead, we can leverage a custom Task Executor introduced in iOS 18 to ensure that the Task's queue pool chooses the internal queue directly, thus  improving performance and minimizing context switching.

### Note:
I have not added a specific Test that validate this workflow, as in, the  Task is executed on the executer's queue. This is because the current test stack in Nuke is large and I could not figure out exactly with the limited time I have, how to leverage current components to write the tests.

I will leave it here, and hopefully, If needed, another, more codebase knowledgeable contributor can implement a test for it.


kean:
Hey, that's a excellent suggestion. It pained me to add this extra hop for the sake of supporting Swift Concurrency.

I have a [separate branch](https://github.com/kean/Nuke/pull/802) where these classes are completely redesigned using actors. It should remove this extra thread hop and a few minor ones during processing. I'm planning to release a beta sometime next month. With that in mind, I don't think it's worth adding something to Nuke 12, but I haven't fully investigated it.

PR #819, opened 2024-10-17
Update SHA1 to SHA256

myersjac:
Updating the Data Cache to utilize a more up-to-date with security standards hashing algorithm.

Tests have also been updated to utilize and confirm the functionality. All Tests passing

As an employee of a large corporation, I would love to utilize this library to load images as it is very performant. Unfortunately, based on our corporate security standards, we are unable to use this framework as it stands because it contains usage of an insecure hash function. This PR's goal is to update that hash function from SHA1 to SHA256, allowing it to adhere to current security standards and will make it so that it does not get flagged as a vulnerability within static scans of the framework

### DIFF
--- a/Sources/Nuke/Internal/Extensions.swift
+++ b/Sources/Nuke/Internal/Extensions.swift
@@ -9,7 +9,7 @@ extension String {
     /// Calculates SHA256 from the given string and returns its hex representation.
     ///
     /// ```swift
-    /// print("http://test.com".256)
+    /// print("http://test.com".sha256)
     /// // prints "8b408a0c7163fdfff06ced3e80d7d2b3acd9db900905c4783c28295b8c996165"
     /// ```
     var sha256: String? {


myersjac:
Screenshot 2024-10-17 at 3 48 11 PM

Sources/Nuke/Internal/Extensions.swift
```diff
@@ -6,17 +6,17 @@ import Foundation
 import CryptoKit
 
 extension String {
-    /// Calculates SHA1 from the given string and returns its hex representation.
+    /// Calculates SHA256 from the given string and returns its hex representation.
     ///
     /// ```swift
-    /// print("http://test.com".sha1)
-    /// // prints "50334ee0b51600df6397ce93ceed4728c37fee4e"
+    /// print("http://test.com".256)
```
duguyihou, line 12: should it be `/// print("http://test.com".sha256)`?

myersjac: oof good catch. let me revise

fabianfabian:
Is there a performance difference when switching to sha256? How big?

Is the hash even used for securing anything here? Those types of corporate policies are usually for password hashing.

kean:
Hey, I just wanted to say that I saw the PR, and I share the same concerns as @fabianfabian. I begrudgingly switched from md5 to (slower) sha1 in one of the previous releases, and sha256 will likely be slower still.

> Is the hash even used for securing anything here?

Absolutely not, but this type of tooling typically doesn't care.

I've been thinking about switching to an SQLite-based cache that wouldn't require hashing.

myersjac:
Yeah, from what I understood from the code itself there aren't any actual security concerns with this.

It's only an issue for me as because it's included within the framework, and that gets picked up in the static scan that we run for security vulnerabilities. 

I believe our use case doesn't utilize this part of the code even, but our company's security policy is really strict that the inclusion of using SHA1 for something is not acceptable.

We would love to use this framework as it's been extremely performant, but as it stands we can't because of the use of SHA1. 

Regardless, do whatever you think is best for it. 👍

PR #822, opened 2024-11-07
Mark DataLoading completion closures as Sendable

ecito:
I know you're focusing on Nuke 13, but this little change would help us out a lot without any downsides.

We implement our own `DataLoading` loader using async/await but we can't use the `DataLoading` closures in a `Task` due to them not being Sendable.

🙏


kean:
Yeah, I'm releasing it soon with this and other (breaking) changes.

PR #839, opened 2025-03-10
Updating misleading SVG support

realmtai:
Nuke does not render SVG out of the box, let's make sure we don't mislead people on this


kean:
Hey, good point. I'd just drop it from this list.

realmtai:
yeap, SVG nuke'd 😉

PR #844, opened 2025-04-06
Start ImageTask in .suspended state

kean:


Sources/Nuke/ImageTask.swift
```diff
@@ -152,26 +151,33 @@ public final class ImageTask: Hashable {
         self.isDataTask = isDataTask
         self.pipeline = pipeline
         self.onEvent = onEvent
-        // TODO: remove thread hop
-        self.task = Task { @ImagePipelineActor in
```
kean, line 156: `ImageTask` is created from a `nonisolated` method (by design), which previously required to spawn a `Task` to perform it. By delaying the task start until you call once of its `async` APIs such as `response: ImageResponse`, we can take advantage of the asynchronous context created by however invokes it. There is no need to spawn one of our own. This eliminates a thread hop.

Sources/Nuke/ImageTask.swift
```diff
@@ -60,16 +60,18 @@ public final class ImageTask: Hashable {
...
     public enum State: Sendable {
+        /// The initial state.
+        case suspended
         /// The task is currently running.
         case running
         /// The task has received a cancel message.
         case cancelled
         /// The task has completed (without being canceled).
-        case completed
+        case completed(Result<ImageResponse, Error>)
```
kean, line 74: It no longer rely on `Task` caching of results.

PR #846, opened 2025-04-10
Fix bug where requests with processors would not check disk cache for original image data

JValldejuli:
`TaskFetchOriginalData` will now check the disk cache for the original image data before making a network request.

### DIFF
--- a/Sources/Nuke/Tasks/TaskFetchOriginalData.swift
+++ b/Sources/Nuke/Tasks/TaskFetchOriginalData.swift
@@ -29,6 +29,10 @@ final class TaskFetchOriginalData: AsyncPipelineTask<(Data, URLResponse?)>, @unc
             return
         }
 
+        if let data = pipeline.cache.cachedData(for: makeSanitizedRequest()) {
+            send(value: (data, nil), isCompleted: true)
+        }
+
         if let rateLimiter = pipeline.rateLimiter {
             // Rate limiter is synchronized on pipeline's queue. Delayed work is
             // executed asynchronously also on the same queue.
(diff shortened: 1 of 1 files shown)


kean:
Hey, thanks for raising the issue and for the extended details. I'm working on redesigning and simplifying the tasks for [Nuke 13](https://github.com/kean/Nuke/pull/802) – I'll address this as part of the rework.

Sources/Nuke/Tasks/TaskFetchOriginalData.swift
```diff
@@ -29,6 +29,10 @@ final class TaskFetchOriginalData: AsyncPipelineTask<(Data, URLResponse?)>, @unc
             return
         }
 
+        if let data = pipeline.cache.cachedData(for: makeSanitizedRequest()) {
```
timothy-barnard, line 32: Would this be better done in the `TaskLoadData` since the `returnCacheDataDontLoad` would block this.

kean:
Thanks for the PR! I think the existing architecture already handles this case. `TaskLoadImage` recursively peels off processors one at a time, and each level checks the disk cache:

```swift
  TaskLoadImage([A, B]) → cache check with key "url+A+B" (miss)
    → TaskLoadImage([A]) → cache check with key "url+A" (miss)
      → TaskLoadImage([]) → cache check with key "url" (hit)
```

Since `storeDataInCacheIfNeeded` stores original data under a sanitized key (no processors), the cache is hit naturally when the processors are fully stripped — see `TaskLoadImage.swift` and the recursive call in `fetchImage()` at lines 52–54.

Could you provide a test case that reproduces the issue? That would help me understand the specific scenario where the cache is being missed.

PR #879, opened 2026-07-10
Add asynchronous image decoding support

### LINKED ISSUE #860: Asynchronous image decoding

thliu21:
## Summary

- add `AsyncImageDecoding` for decoders that need to call asynchronous APIs
- await async decoders on the existing image decoding queue while preserving error wrapping, cancellation, priority, and signpost behavior
- keep existing synchronous `ImageDecoding` implementations source-compatible
- document async decoder registration and the synchronous cache API limitation
- add coverage for successful async decoding and error propagation

## Motivation

`ImageDecoding` currently exposes only a synchronous entry point, so implementations backed by async APIs have to block a thread with a semaphore. This adds a dedicated async refinement that the pipeline can detect and await without changing existing decoders.

Closes #860.

## Testing

- `swift build` (all package targets)
- `ImagePipelineDecodingTests`: 5 passed
- `ImagePipelineTests` and `ImagePipelineProgressiveDecodingTests`: 35 passed
- DocC build

The macOS test runs used Thread Sanitizer disabled because the local Xcode 26.3 test runner could not load its TSan runtime; the shared scheme remains unchanged.

### DIFF
--- a/Documentation/Nuke.docc/Customization/ImageFormats/image-decoding.md
+++ b/Documentation/Nuke.docc/Customization/ImageFormats/image-decoding.md
@@ -23,25 +23,6 @@ public protocol ImageDecoding: Sendable {
 }
 ```
 
-If decoding needs to call asynchronous APIs, use ``AsyncImageDecoding`` instead:
-
-```swift
-final class CustomDecoder: AsyncImageDecoding {
-    func decode(_ data: Data) async throws -> ImageContainer {
-        try await CustomDecoderImplementation.decode(data)
-    }
-}
-```
-
-The pipeline recognizes async decoders returned by `makeImageDecoder` or
-``ImageDecoderRegistry`` and awaits them on the image decoding queue. There is
-no need to block a thread with a semaphore. Progressive decoding remains
-synchronous and can be implemented with `decodePartiallyDownloadedData(_:)`.
-
-> Note: The synchronous ``ImagePipeline/Cache/cachedImage(for:caches:)`` API
-> can't use an async-only decoder to decode disk data. Regular image pipeline
-> loading, including loading from the disk cache, supports async decoders.
-
 ``ImageContainer`` is a struct that wraps the decoded image and (optionally) the original data and some additional information. The decoder decides what to attach to the container.
 
 ```swift
--- a/Documentation/Nuke.docc/Customization/ImageFormats/image-decoding.md
+++ b/Documentation/Nuke.docc/Customization/ImageFormats/image-decoding.md
@@ -23,6 +23,25 @@ public protocol ImageDecoding: Sendable {
 }
 ```
 
+If decoding needs to call asynchronous APIs, use ``AsyncImageDecoding`` instead:
+
+```swift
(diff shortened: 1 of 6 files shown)


kean:
Hey, thanks for the PR! This is a problem worth solving, and I've been thinking about it on and off.

A few thoughts in no particular order:

- Introducing a new `AsyncImageDecoding` protocol is a good option: it's a relatively niche use case, and this approach keeps the change non-breaking.
- There are scenarios it doesn't cover, such as [ImagePipeline.Cache.cachedImage(for:caches:)](https://github.com/kean/Nuke/blob/10a9b48565c5f0c2851e4462ac877d02b7154f59/Sources/Nuke/Pipeline/ImagePipeline%2BCache.swift#L52), which relies on decoding being synchronous.
- As a workaround, you can currently use semaphores. It's not great, but not terrible.

I'm also considering a more comprehensive change: making the base `ImageDecoding` protocol async, along with `ImagePipeline.Cache.cachedImage(for:caches:)` and other APIs like `ImageProcessing`, `DataLoading`, and more. That's what you'd expect from a modern framework. One thing I'm still on the fence about is whether to isolate these APIs to `ImagePipelineActor` to avoid paying for the thread hops. With Nuke 13 released almost half a year ago, now is probably a good time to start working on a new major version that includes this breaking change.

_Update_: thread hops in the case of `ImageDecoding` is less of a concern because the internal `TaskQueue` already supports `async` and uses it for all work.

thliu21:
Thanks — this matches the trade-off I was aiming for. `cachedImage(for:caches:)` is the known gap: because it is synchronous, an async-only decoder can’t participate in disk-cache decoding there without blocking. The PR calls this out in the decoding docs, while regular pipeline loads from the disk cache do await `AsyncImageDecoding`.

Your `TaskQueue` update also aligns with the implementation here: async decoding is enqueued directly on `imageDecodingQueue`, without introducing a separate background hop.

Given the broader next-major direction, would you prefer this PR as an interim non-breaking API—public or behind `@_spi`—or should we hold it and fold the work into making the base protocols async? I’m happy to adjust either way.

kean:
> @_spi

This is probably ideal. I keep forgetting about. I'm open to merging and shipping it with `@_spi`. I'd love to provide an escape hatch, but without "officially" supporting it just yet.

PR #880, opened 2026-07-15
Fix GaussianBlur crash on grayscale images

Delarkz:
`ImageProcessors.GaussianBlur` crashes (SIGSEGV) on grayscale (monochrome) images like black-and-white JPEGs. It shows up on iOS and tvOS; on macOS the same out-of-bounds write happens but usually corrupts memory silently instead of faulting.

`vImageBoxConvolve_ARGB8888` assumes 4 bytes per pixel, but the scratch contexts were created with the source's own color space. For a grayscale image that's a 16-bit gray+alpha buffer (2 bytes per pixel), so vImage ends up reading and writing past the end of the allocation.

The fix creates the scratch contexts in device RGB so the buffer layout matches what vImage expects (drawing a gray image into an RGB context is lossless).


kean:
Nice catch, thank you!

PR #881, opened 2026-07-31
Fix cache initialization for mergeable libraries

### LINKED ISSUE #826: ImagePipeline init crashing when using Mergeable libraries
After switching to [mergeable libraries](https://developer.apple.com/documentation/xcode/configuring-your-project-to-use-mergeable-libraries) Nuke is crashing. Upon debugging below is the trace which I get in Xcode organiser 
Screenshot 2024-12-24 at 12 06 28 PM

My probable guess is memoryPressure was in suspended state, this could be easily fixed by replacing `memoryPressure.resume()` with `memorySource.activate()`

ZHUOLIN0928:
## Summary

- initialize the memory-pressure dispatch source with `activate()` when Nuke is used with mergeable libraries
- avoid resuming an already active source during cache initialization

Fixes #826

## Validation

- `swift build --target Nuke` ✅
- `swift test --filter NukeTests` could not complete locally because the installed CommandLineTools cannot resolve the existing `SwiftUIMacros.StateMacro` plugin while compiling NukeUI; this is unrelated to the changed file

### DIFF
--- a/Sources/Nuke/Caching/Cache.swift
+++ b/Sources/Nuke/Caching/Cache.swift
@@ -79,7 +79,7 @@ final class Cache<Key: Hashable & Sendable, Value: Sendable>: @unchecked Sendabl
         self.memoryPressure.setEventHandler { [weak self] in
             self?.removeAllCachedValues()
         }
-        self.memoryPressure.resume()
+        self.memoryPressure.activate()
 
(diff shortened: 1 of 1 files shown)


kean:
TIL `activate`. Thank you!

PR #884, opened 2026-08-11
Fix isResetEnabled progressive image loading

### LINKED ISSUE #883: Progressive Image Downloads Fail to Complete if `isResetEnabled = true` when the request starts
If `isResetEnabled = true` when the request starts, `resetOrDefer` will defer resetting by raising `isResetNeeded`.

However, if `isProgressiveImageRenderingEnabled = true`, `display` is called to display the partial image, but since resetting was deferred, `resetIfNeeded` now resets which cancels the in progress image download

nathantannar4:
Fixes #883, progressive image download canceled if `isResetEnabled == true` when request is started

### DIFF
--- a/Sources/NukeUI/LazyImageView.swift
+++ b/Sources/NukeUI/LazyImageView.swift
@@ -242,8 +242,10 @@ public final class LazyImageView: _PlatformBaseView {
         reset(clearImage: true)
     }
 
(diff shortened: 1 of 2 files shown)


kean (approved):
Thanks for the fix, and the test!

PR #889, opened 2026-08-15
Fix disk cache key collisions

kean:
`ImageProcessors.Composition.identifier` and `makeDataCacheKey(for:)` concatenated identifiers with no separator, so requests like `["resize-", "largeblur"]` and `["resize-large", "blur"]` produced the same disk cache key and the second one was served the first one's processed bytes. The same applied to the `imageID`/`thumbnail`/processors boundaries. The memory cache didn't have this bug, so the two cache layers disagreed.

Both now join their components with a `+` separator (always emitted, so empty components can't alias either).

Note: this changes the on-disk cache key format, so existing disk cache entries become unreachable - they simply miss, get re-fetched, and are reclaimed by the normal `DataCache` sweep.


kean:
This is a breaking change. I'd need a migration step to drop the current cache.

PR #890, opened 2026-08-15
Deliver ImageTask.Event.started through the events stream

kean:
`startImageTask` sent `.started` straight to the pipeline delegate and never called `_dispatch`, which is the only feeder of the public `events` stream and of the `onEvent` closure. `for await event in task.events { if case .started … }` could therefore never fire, even though `.started` is a public case of the stream's element type and `events` is documented as "the events sent by the pipeline during the task execution". `Deprecated.swift`'s `case .started: break` was dead code.

The event now goes through `_dispatch` like every other one. The pipeline starts the task asynchronously, so a subscriber has no way to be registered before `.started` is sent — it is also replayed to the streams created afterwards, which makes it the first event of every stream.

`.started` is now also delivered on the `events` stream of the data tasks, matching `.progress` and `.finished`. The delegate callback is unchanged.

### DIFF
--- a/Sources/Nuke/ImageTask.swift
+++ b/Sources/Nuke/ImageTask.swift
@@ -131,7 +131,9 @@ public final class ImageTask: Hashable, CustomStringConvertible, @unchecked Send
 
     /// An event produced during the runtime of the task.
     @frozen public enum Event: Sendable {
-        /// The task was started by the pipeline.
+        /// The task was started by the pipeline. Always the first event in the
+        /// stream: it is replayed for the subscribers that are added after the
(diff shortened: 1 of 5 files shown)


kean:
There is really no point in this event considering that the tasks are eager. It was initially added as a replacement for per-vent-type method in `ImagePipeline.Delegate`. It's not really meant to be sent in first-class events in `ImageTask` event-producing properties/methods.

kean:
Replaced it with a dedicated method https://github.com/kean/Nuke/pull/914. The stream should NOT have this event.

PR #938, opened 2026-08-29
Rename Border.width to widthInPixels

kean:
`ImageProcessingOptions.Border` takes a width plus a unit and stores the converted value, but reported it through a property named `width`, so it never returned what the caller passed:

```swift
Border(color: .red, width: 2, unit: .points).width // 6 on a 3x screen, not 2
```

The stored property is now `widthInPixels`, which is what `description` already printed. `width` stays as a deprecated alias for the same value.


kean:
Let's keep it as is for this iteration.

PR #959, opened 2026-09-02
Run queued operations at the priority they were scheduled with

DePasqualeOrg:
This resolves an Xcode warning about priority inversion.

### DIFF
--- a/Sources/Nuke/Internal/Extensions.swift
+++ b/Sources/Nuke/Internal/Extensions.swift
@@ -50,6 +50,17 @@ extension ImageRequest.Priority {
     }
 }
 
+extension TaskPriority {
+    var concurrencyPriority: _Concurrency.TaskPriority {
+        switch self {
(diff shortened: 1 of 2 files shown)


kean:
> This resolves an Xcode warning about priority inversion.

Hey, @DePasqualeOrg, would mind sharing the warning? I'm not sure I've seen one.

DePasqualeOrg:
Xcode highlighted `CGImageSourceCreateThumbnailAtIndex` inside Nuke’s `makeThumbnail` while a `LazyImage` loaded an uncached image with `ImageRequest.ThumbnailOptions`. I believe this was caused by the missing task priority in `TaskQueue.execute`. This was the Xcode warning:

> [Internal] Thread running at User-initiated quality-of-service class waiting on a lower QoS thread running at Default quality-of-service class. Investigate ways to avoid priority inversions

